PR #3, opened 2016-01-20
Draft PEP

rmcgibbo:
This is basically @njsmith's synopsis from README.rst reformatted as a PEP.

### DIFF
--- a/pep-XXXX.rst
+++ b/pep-XXXX.rst
@@ -6,15 +6,15 @@ Author: Robert T. McGibbon <rmcgibbo@gmail.com>, Nathaniel J. Smith <njs@pobox.c
 Status: Draft
 Type: Process
 Content-Type: text/x-rst
-Created: 01-Jan-2016
-Post-History: 30-Aug-2002
+Created: 19-Jan-2016
+Post-History: 19-Jan-2016
 
 
 Abstract
 ========
 
 This PEP proposes the creation of a new platform tag for Python package built
-distributions, such as wheels, called ``manylinux_1_{x86_64,i386}`` with
+distributions, such as wheels, called ``manylinux1_{x86_64,i386}`` with
 external dependencies limited restricted to a standardized subset of
 the Linux kernel and core userspace ABI. It proposes that PyPI support
 uploading and distributing Wheels with this platform tag, and that ``pip``
@@ -56,9 +56,9 @@ extension modules for Linux -- e.g. Enthought with Canopy [2]_ and Continuum
 Analytics with Anaconda [3]_.
 
 Building on the compability lessons learned from these companies, we thus
-define a baseline ``manylinux_1`` platform tag for use by binary Python
+define a baseline ``manylinux1`` platform tag for use by binary Python
 wheels, and introduce the implementation of preliminary tools to aid in the
-construction of these ``manylinux_1`` wheels.
+construction of these ``manylinux1`` wheels.
 
 
 Key Causes of Inter-Linux Binary Incompatibility
@@ -110,8 +110,8 @@ This generally prevents built distributions compiled on the latest Linux
 distributions from being portable.
 
 
-The ``manylinux_1`` policy
-==========================
+The ``manylinux1`` policy
+=========================
 
 For these reasons, to achieve broad portability, Python wheels
 
@@ -120,7 +120,7 @@ For these reasons, to achieve broad portability, Python wheels
  * should depend only on ``old`` symbol versions in those external shared
    libraries.
 
-The ``manylinux_1`` policy thus encompasses a standard for what the
+The ``manylinux1`` policy thus encompasses a standard for what the
 permitted external shared libraries a wheel may depend on, and the maximum
 depended-upon symbol versions therein.
 
@@ -166,11 +166,11 @@ this date.
 Compilation and Tooling
 =======================
 
-To support the compilation of wheels meeting the ``manylinux_1`` standard, we
+To support the compilation of wheels meeting the ``manylinux1`` standard, we
 provide initial drafts of two tools.
 
 The first is a Docker image based on CentOS 5.11, which is recommended as an
-easy to use self-contained build box for compiling  ``manylinux_1`` wheels.
+easy to use self-contained build box for compiling  ``manylinux1`` wheels.
 Compiling on a more recently-released linux distribution will generally
 introduce dependencies on too-new versioned symbols. The image comes with a
 full compiler suite installed (``gcc``, ``g++``, and ``gfortran`` 4.8.2) as
@@ -179,7 +179,7 @@ well as the latest releases of Python and pip.
 The second tool is a command line executable called ``auditwheel``. First, it
 inspects all of the ELF files inside a wheel to check for dependencies on
 versioned symbols or external shared libraries, and verifies conformance with
-the ``manylinux_1`` policy. This includes the ability to add the new platform
+the ``manylinux1`` policy. This includes the ability to add the new platform
 tag to conforming wheels.
 
 In addition, ``auditwheel`` has the ability to automatically modify wheels that
@@ -190,7 +190,7 @@ similar result as if the libraries had been statically linked without requiring
 changes to the build system.
 
 Neither of these tools are necessary to build wheels which conform with the
-``manylinux_1`` policy. Similar results can usually be achieved by statically
+``manylinux1`` policy. Similar results can usually be achieved by statically
 linking external dependencies and/or using certain inline assembly constructs
 to instruct the linker to prefer older symbol versions, however these tricks
 can be quite esoteric.
@@ -199,7 +199,7 @@ can be quite esoteric.
 Platform Detection in ``pip``
 =============================
 
-TODO How does ``pip`` detect that it's running on a ``manylinux_1`` compatible
+TODO How does ``pip`` detect that it's running on a ``manylinux1`` compatible
 system?
 
 
@@ -213,7 +213,7 @@ effects of these patches when the underlying libraries are updated. This can
 be particularly important for security updates in packages communication
 across the network or cryptography.
 
-``manylinux_1`` wheels distributed through PyPI that bundle security-critical
+``manylinux1`` wheels distributed through PyPI that bundle security-critical
 libraries like OpenSSL will thus assume responsibility for prompt updates in
 response disclosed vulnerabilities and patches. This closely parallels the
 security implications of the distribution of binary wheels on Windows that,
--- a/pep-XXXX.rst
+++ b/pep-XXXX.rst
@@ -211,11 +211,74 @@ to instruct the linker to prefer older symbol versions, however these tricks
 can be quite esoteric.
 
 
-Platform Detection in ``pip``
-=============================
+Platform Detection for Installers
+=================================
 
-TODO How does ``pip`` detect that it's running on a ``manylinux1`` compatible
-system?
+Because the ``manylinux1`` profile is already known to work for the many
+thousands of users of popular commercial Python distributions, we suggest that
+installation tools like ``pip`` should error on the side of assuming that a
+system *is* compatible, unless there is specific reason to think otherwise.
+
+We know of three main sources of potential incompatibility that are likely to
+arise in practice:
+
+* A linux distribution that is too old (e.g. RHEL 4)
+* A linux distribution that does not use glibc (e.g. Alpine Linux, which is
+  based on musl libc, or Android)
+* Eventually, in the future, there may exist distributions that break
+  compatibility with this profile
+
+To handle the first two cases, we propose the following simple and reliable
+check: ::
+
+    def have_glibc_version(major, minimum_minor):
+        import ctypes
+
+        process_namespace = ctypes.CDLL(None)
+        try:
+            gnu_get_libc_version = process_namespace.gnu_get_libc_version
+        except AttributeError:
+            # We are not linked to glibc.
+            return False
+
+        gnu_get_libc_version.restype = ctypes.c_char_p
+        version_str = gnu_get_libc_version()
+        # py2 / py3 compatibility:
+        if not isinstance(version_str, str):
+            version_str = version_str.decode("ascii")
+
+        version = [int(piece) for piece in version_str.split(".")]
+        assert len(version) == 2
+        if major != version[0]:
+            return False
+        if minimum_minor > version[1]:
+            return False
+        return True
+
+    # CentOS 5 uses glibc 2.5.
+    is_manylinux1_compatible = have_glibc_version(2, 5)
+
+To handle the third case, we propose the creation of a file
+``/etc/python/compatibility.cfg`` in ConfigParser format, with sample
+contents: ::
+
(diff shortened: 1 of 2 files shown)


pep-XXXX.rst
```diff
@@ -0,0 +1,258 @@
+PEP: XXXX
+Title: A Platform Tag for Portable Linux Built Distributions
+Version: $Revision$
+Last-Modified: $Date$
+Author: Robert T. McGibbon <rmcgibbo@gmail.com>, Nathaniel J. Smith <njs@pobox.com>
+Status: Draft
+Type: Process
+Content-Type: text/x-rst
+Created: 01-Jan-2016
+Post-History: 30-Aug-2002
```
njsmith: Pst, your time machine is showing ;-)

pep-XXXX.rst
```diff
@@ -0,0 +1,258 @@
...
+
+
+Abstract
+========
+
+This PEP proposes the creation of a new platform tag for Python package built
+distributions, such as wheels, called ``manylinux_1_{x86_64,i386}`` with
```
njsmith: Okay, here's a pure style/taste nitpick: should it be `manylinux_1` or `manylinux1`? I prefer `manylinux1` just for consistency with the look of most of the other wheel tags `py2`, `cp35`, `win32`, `i386`, `amd64` (OTOH there's also `macosx_10_9` and `x86_64`, but I think they're the only ones, so I guess the pattern is that underscores before numbers are only used when multiple numbers go together?). Doesn't matter too much, but probably worth a few seconds of thought.

rmcgibbo:
in terms of the outline, do you think this hits the proper points? i've never written one of these before.

some of my concerns are:

- how much should we talk about the tools (docker image, auditwheel)? should we include example command-line invocations for compiling some packages, for example?
- the required changes to PyPI should probably be made more explicit.
- we should include a paragraph about the anticipated procedure for updating the whitelist (presumably to CentOS 6).
- we should probably fork pip into this organization to draft the changes required in pip for the platform detection. i'm not exactly sure how this is supposed to work (you mentioned a file in `/etc` for example).

pep-XXXX.rst
rmcgibbo: Okay, that seems convincing.

pep-XXXX.rst
```diff
@@ -0,0 +1,258 @@
...
+depended-upon symbol versions therein.
+
+The permitted external shared libraries are: ::
+
+  "libpanelw.so.5", "libncursesw.so.5", "libgcc_s.so.1", "libstdc++.so.6",
+  "libm.so.6", "libdl.so.2", "librt.so.1", "libcrypt.so.1", "libc.so.6",
+  "libnsl.so.1", "libutil.so.1", "libpthread.so.0", "libX11.so.6",
+  "libXext.so.6", "libXrender.so.1", "libICE.so.6", "libSM.so.6",
+  "libGL.so.1", "libgobject-2.0.so.0", "libgthread-2.0.so.0",
+  "libglib-2.0.so.0", "libgmodule-2.0.so.0", and "libgio-2.0.so.0".
```
njsmith: 1) Can we format these as big vertical lists like
```
  libpanelw.so.5
  libncursesw.so.5
  ...
```
etc. without the quotes and commas? It's way more friendly for cut-and-pasting that way, and we're not going to run out of vertical space :-)

2) Do we really want to include `libgmodule-2.0.so.0` and `libgio-2.0.so.0`? The way they get thrown in at the end is triggering the same instinct that makes me twitch when I see "one little added feature" thrown into a final release after being left out of the betas, because *surely* it won't cause any problems :-). The fact that these seem to be almost or totally unused by Anaconda/Canopy suggests that no-one will miss them if we leave them out. Also we don't actually know whether they'll work (they'll probably be present, but for all we know libgio has broken ABI recently, we just don't have data). And also it's a lot easier rhetorically if we can say "everything in this list is used by both Anaconda and Enthought", versus "everything in this list is used by both Anaconda and Enthought [except ...]"?

pep-XXXX.rst
```diff
@@ -0,0 +1,258 @@
...
+of the most popular Python modules and have been confirmed in practice to work
+across a wide swath of Linux systems in the wild.
+
+For dependencies on externally-provided versioned symbols in the above shared
+libraries, the following symbol versions are permitted: ::
+
+    GLIBC <= 2.5
+    CXXABI <= 3.4.8
+    GLIBCXX <= 3.4.9
+    GCC <= 4.2.0
```
njsmith: Do we want to mention ncurses symbol versions here?

pep-XXXX.rst
rmcgibbo: 1. Sure.

2. libgmodule-2.0.so.0 is in https://github.com/manylinux/manylinux/blob/master/policy-info/canopy-1.6.1-libraries-needed-for-main-env though?

pep-XXXX.rst
rmcgibbo: I tried to look (`readelf -V /usr/lib/libncurses.so`) and I didn't see any ncurses symbol versions. How did you find it?

njsmith: Weird, on my (debian testing) system I see tons of symbol versions in there: https://gist.github.com/njsmith/06b60d6a8416250b60c6

njsmith: (That gist is for libncursesw since that's the version in the current profile, but I see similar results for libncurses-without-the-w, so that's not the issue)

rmcgibbo: on centos5.11, there are none of those symbols.

njsmith: Heh, you responded faster than I could download the image from quay.io to check :-)

pep-XXXX.rst
njsmith: Re: libgmodule: yeah -- looking at the full list in canopy-further-investigation it appears to be used only by wx-gtk (which also depends on a bunch of even heavier stuff like gtk), and there's lots of other stuff on that list that we don't have in the policy (gtk itself, libpng, GLU...). So my inclination is to [draw a simple, defensible, objective line line](https://groups.google.com/d/msg/manylinux-discuss/NR-TN_g5gJk/U7mCeShhFgAJ) rather than try to come up with exactly the right answer for every library? But I could be argued into gmodule, I guess...

rmcgibbo: Kk. I'll take it out.

pep-XXXX.rst
rmcgibbo: Perhaps I should just say that no other symbol version name other than the explicitly listed ones are permitted.

pep-XXXX.rst
```diff
@@ -0,0 +1,258 @@
...
+linking external dependencies and/or using certain inline assembly constructs
+to instruct the linker to prefer older symbol versions, however these tricks
+can be quite esoteric.
+
+
+Platform Detection in ``pip``
+=============================
+
+TODO How does ``pip`` detect that it's running on a ``manylinux1`` compatible
+system?
```
njsmith: Suggest text:

```
Because the ``manylinux1`` profile is already known to work for the many thousands of users of popular commercial Python distributions, we suggest that installation tools like ``pip`` should error on the side of assuming that a system *is* compatible, unless there is specific reason to think otherwise.

We know of three main sources of potential incompatibility that are likely to arise in practice:

* A linux distribution that is too old (e.g. RHEL 4)
* A linux distribution that does not use glibc (e.g. Alpine Linux, which is based on musl libc, or Android)
* Eventually, in the future, there may exist distributions that break compatibility with this profile

To handle the first two cases, we propose the following simple and reliable check::

    def have_glibc_version(major, minimum_minor):
        import ctypes

        process_namespace = ctypes.CDLL(None)
        try:
            gnu_get_libc_version = process_namespace.gnu_get_libc_version
        except AttributeError:
            # We are not linked to glibc.
            return False

        gnu_get_libc_version.restype = ctypes.c_char_p
        version_str = gnu_get_libc_version()
        # py2 / py3 compatibility:
        if not isinstance(version_str, str):
            version_str = version_str.decode("ascii")

        version = [int(piece) for piece in version_str.split(".")]
        assert len(version) == 2
        if major != version[0]:
            return False
        if minimum_minor > version[1]:
            return False
        return True

    # CentOS 5 uses glibc 2.5.
    is_manylinux1_compatible = have_glibc_version(2, 5)

To handle the third case, we propose the creation of a file `/etc/python/compatibility.cfg` in ConfigParser format, with sample contents::

   [manylinux1]
   compatible = true

where the supported values for the manylinux1.compatible entry are the same as those supported by the ConfigParser getboolean method.

The proposed logic for ``pip`` or related tools, then, is:

0) If ``distutils.util.get_platform()`` does not start with the string ``"linux"``, then assume the current system is not ``manylinux1`` compatible.
1) If /etc/python/compatibility.conf exists and contains a ``manylinux1`` key, then trust that.
2) Otherwise, if ``have_glibc_version(2, 5)`` returns true, then assume the current system can handle ``manylinux1`` wheels.
3) Otherwise, assume that the current system cannot handle ``manylinux1`` wheels.
```

pep-XXXX.rst
```diff
@@ -0,0 +1,258 @@
...
+effects of these patches when the underlying libraries are updated. This can
+be particularly important for security updates in packages communication
+across the network or cryptography.
+
+``manylinux1`` wheels distributed through PyPI that bundle security-critical
+libraries like OpenSSL will thus assume responsibility for prompt updates in
+response disclosed vulnerabilities and patches. This closely parallels the
+security implications of the distribution of binary wheels on Windows that,
+because the platform lacks a system package manager, generally bundle their
+dependencies.
```
njsmith: We might want to note here explicitly that openssl in particular does not have a stable ABI and so cannot be included in the `manylinux1` profile .

pep-XXXX.rst
rmcgibbo: is the file really supposed to be in `/etc/python/compatibility.cfg` in the real root `/etc`, even for a userspace installation of python/pip?

njsmith: Huh, intersting question that I don't actually know the best answer to :-)
might be something to ask distutils-sig about. (Nick Coghlan will probably
have an informed opinion if nothing else.)

On Wed, Jan 20, 2016 at 5:01 PM, Robert T. McGibbon <

> In pep-XXXX.rst
> <https://github.com/manylinux/manylinux/pull/3#discussion_r50346463>:
>
> > +such that these libraries will be picked up at runtime. This accomplishes a
> > +similar result as if the libraries had been statically linked without requiring
> > +changes to the build system.
> > +
> > +Neither of these tools are necessary to build wheels which conform with the
> > +``manylinux1`` policy. Similar results can usually be achieved by statically
> > +linking external dependencies and/or using certain inline assembly constructs
> > +to instruct the linker to prefer older symbol versions, however these tricks
> > +can be quite esoteric.
> > +
> > +
> > +Platform Detection in ``pip``
> > +=============================
> > +
> > +TODO How does ``pip`` detect that it's running on a ``manylinux1`` compatible
> > +system?
>
> is the file really supposed to be in /etc/python/compatibility.cfg in the
> real root /etc, even for a userspace installation of python/pip?
>
> —
> <https://github.com/manylinux/manylinux/pull/3/files#r50346463>.
>



-- 

njsmith: (For the record, the name /etc/python/compatibility.cfg is inspired by
the /etc/python/cert-verification.cfg
in Nick's PEP 493)

On Wed, Jan 20, 2016 at 5:05 PM, Nathaniel Smith <njs@pobox.com> wrote:

> Huh, intersting question that I don't actually know the best answer to :-)
> might be something to ask distutils-sig about. (Nick Coghlan will probably
> have an informed opinion if nothing else.)
>
> On Wed, Jan 20, 2016 at 5:01 PM, Robert T. McGibbon <
> notifications@github.com> wrote:
>
>> In pep-XXXX.rst
>> <https://github.com/manylinux/manylinux/pull/3#discussion_r50346463>:
>>
>> > +such that these libraries will be picked up at runtime. This accomplishes a
>> > +similar result as if the libraries had been statically linked without requiring
>> > +changes to the build system.
>> > +
>> > +Neither of these tools are necessary to build wheels which conform with the
>> > +``manylinux1`` policy. Similar results can usually be achieved by statically
>> > +linking external dependencies and/or using certain inline assembly constructs
>> > +to instruct the linker to prefer older symbol versions, however these tricks
>> > +can be quite esoteric.
>> > +
>> > +
>> > +Platform Detection in ``pip``
>> > +=============================
>> > +
>> > +TODO How does ``pip`` detect that it's running on a ``manylinux1`` compatible
>> > +system?
>>
>> is the file really supposed to be in /etc/python/compatibility.cfg in
>> the real root /etc, even for a userspace installation of python/pip?
>>
>> —
>> Reply to this email directly or view it on GitHub
>> <https://github.com/manylinux/manylinux/pull/3/files#r50346463>.
>>
>
>
>
> --
> Nathaniel J. Smith -- https://vorpus.org <http://vorpus.org>
>



-- 

pep-XXXX.rst
```diff
@@ -0,0 +1,258 @@
...
+
+
+Rejected Alternatives
+=====================
+
+One alternative is provide separate platform tags for each Linux distribution
+(and each version thereof). This would require that package authors would be
+required to compile and upload twenty or more different built distributions of
+their package to PyPI to cover the common linux distributions in use, which we
+consider too onerous to be practical.
```
njsmith: Suggested tweaked text, to avoid unnecessarily tweaking any noses:

One alternative would be to provide separate platform tags for each Linux distribution (and each version thereof), e.g. ``RHEL6``, ``ubuntu14_10``, ``debian_jessie``, etc. Nothing in this proposal rules out the possibility of adding such platform tags in the future, or of further extensions to wheel metadata that would allow wheels to declare dependencies on external system-installed packages. However, such extensions would require substantially more work than this proposal, and still might not be appreciated by package developers who would prefer not to have to maintain multiple build environments and build multiple wheels in order to cover all the common Linux distributions. Therefore we consider such proposals to be out-of-scope for this PEP.

rmcgibbo: yes, this is much better.

njsmith:
Okay, those are my comments :-). I think the next step after that is to email to distutils-sig -- do you want to do that?

rmcgibbo:
let me add the urls for the docker image and auditwheel.

rmcgibbo:
but no, you should email distutils-sig, since i think they know you and trust you more.

njsmith:
The only reason they know and trust me more is because I've emailed them
more often :-) It wouldn't hurt them to get to know more folks from the
scientific python side of things... but if you'd rather not then I can do
it.

On Wed, Jan 20, 2016 at 5:13 PM, Robert T. McGibbon <

> but no, you should email distutils-sig, since i think they know you and
> trust you more.
>
> —
> <https://github.com/manylinux/manylinux/pull/3#issuecomment-173419221>.
>



-- 

rmcgibbo:
Okay. I'm finished with my edits now. You should probably merge the PR before sending the email.

njsmith:
Okay, let's do this!

PR #4, opened 2016-01-20
Make sure docker image contains all the desired devel packages

njsmith:
It wasn't clear to me whether we were catching all these, plus I think we were uninstalling ncurses-devel, so factor out and install them as the last step.


rmcgibbo:
lgtm.

PR #7, opened 2016-01-21
Notes

rmcgibbo:
Here are my notes on the discussion happening on distutils-sig, with an eye towards the changes we might want to make to the wording of the PEP. Overall I'm happy to see such a positive response.

### DIFF
--- a/notes.md
+++ b/notes.md
@@ -59,6 +59,9 @@ Later, he asked:
     development distribution, or do you "trust" people building such wheels?
 
 
+Nathaniel wrote:
 
-
-
+    The rule is basically: "if your wheel
+    works when given access to CentOS 5's versions of the following
+    packages: ..., then your wheel is manylinux1 compatible". Any method
+    for achieving that is fair game :-).
--- a/notes.md
+++ b/notes.md
@@ -65,3 +65,23 @@ Nathaniel wrote:
     works when given access to CentOS 5's versions of the following
     packages: ..., then your wheel is manylinux1 compatible". Any method
     for achieving that is fair game :-).
+
+
+
+
+From Nick :
+
+    The PEP should also be explicit that this does reintroduce the
+    bundling problem that distro unbundling policies were designed to
+    address, but:
+
+    1. In these days of automated continuous intregration & deployment
+       pipelines, publishing new versions and updating dependencies is easier
+       than it was when those policies were defined
+    2. Folks remain free to use "--no-binary" if they want to force local
+       builds rather than using pre-built wheel files
+    3. The popularity of container based deployment and "immutable
+       infrastructure" models involve substantial bundling at the application
+       layer anyway
+    4. This PEP doesn't rule out the idea of offering more targeted
+       binaries for particular Linux distributions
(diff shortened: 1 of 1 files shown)


njsmith:
I'm also wondering if we should rephrase the main section defining the profile to be something like "that 'if it works on CentOS 5 + ... then it's manylinux1, otherwise not'", and then elaborate on what that means. 

One advantage of this is that it would also implicitly specify the kernel ABI version, which on further thought would be a good idea -- IIRC from talking to @cournape it's very rare for python packages to care whether they're built on an old or new kernel, but there are a few exceptions (maybe psutil?) that do care, and therefore have to be built in a real virtual image using an old kernel rather than in a docker container running on a new kernel.

cournape:
@njsmith the only example I am sure causes trouble because of kernel is `subprocess32`

rmcgibbo:
> I'm also wondering if we should rephrase the main section defining the profile to be something like "that 'if it works on CentOS 5 + ... then it's manylinux1, otherwise not'", and then elaborate on what that means.

As in, it needs to both work on a stock CentOS 5 _and_ have at maximum the following shared library dependencies and symbol version dependencies...? That sounds like an improvement in clarity to me.

njsmith:
I guess the way I would phrase it is: it needs to work on stock a CentOS 5 where the only libraries available are the stock CentOS 5 versions of the following: ...

Then the symbol version dependencies etc. are consequences rather than extra restrictions?

rmcgibbo:
The symbol version dependencies are is a necessary consequence of CentOS 5. But the limit on external shared libraries is not -- it is a further limit above and beyond "mere" CentOS 5.

njsmith:
Right

On Thu, Jan 21, 2016 at 3:55 PM, Robert T. McGibbon <

> The symbol version dependencies are is a necessary consequence of CentOS
> 5. But the limit on external shared libraries is not -- it is a further
> limit above and beyond "mere" CentOS 5.
>
> —
> <https://github.com/manylinux/manylinux/pull/7#issuecomment-173753712>.
>



-- 

rmcgibbo:
The usefulness of the explicit symbol policy is that it facilitates checking for auditwheel, and it also points out, if you're into things like custom glibcs or inline assembly, how to compile manylinux1-compatible wheels on non-CentOS 5 platforms.

But +1 to adding the language you mentioned. It will make the standard more explicit.

njsmith:
* [ ] the shared/static libpython issue should get discussed better

rmcgibbo:
What is there to discuss?

notes.md
```diff
@@ -0,0 +1,87 @@
...
+    for achieving that is fair game :-).
+
+
+
+
+From Nick :
+
+    The PEP should also be explicit that this does reintroduce the
+    bundling problem that distro unbundling policies were designed to
+    address, but:
```
rmcgibbo: Does anyone know where to find a link to the formal unbundling policy documents from and major Linux distros? I found Fedora's policy here: https://fedoraproject.org/wiki/Bundled_Software_policy, but it would be nice to be able to cite Debian or something when discussing this issue too.

njsmith: https://wiki.debian.org/EmbeddedCodeCopies
https://www.debian.org/doc/debian-policy/ch-source.html#s-embeddedfiles

njsmith:
I just meant the confusion someone had about whether shared libpython was allowed -- we should just make sure that it's clear that it is allowed :-)

PR #9, opened 2016-01-22
Compile Python with latest OpenSSL

rmcgibbo:


njsmith:
Awesome. (And nice one on the use of sha256sum.)

Is there any easy way to test this, do you suppose?

Can you think of any way to reduce the likelihood that wheels will pick up our version of openssl? I would hate to be responsible for a cve because some project carefully built the latest openssl, but then accidentally bundled our version that we had forgotten to update because auditwheel found the wrong one...

rmcgibbo:
I'm testing it now, and trying to get python statically linked to openssl, so then I can delete `libssl.a` afterwards :)

rmcgibbo:
This works nicely now. I've built the image locally and played around with it, and queried `https://www.howsmyssl.com/` from Python, and everything is kosher.

njsmith:
that's so mysterious. you're using a non-standard install location. HOW DOES PYTHON KNOW.

rmcgibbo:
> that's so mysterious. you're using a non-standard install location. HOW DOES PYTHON KNOW.

?

rmcgibbo:
oh, you mean `/usr/local/ssl`?

rmcgibbo:
https://github.com/python/cpython/blob/master/setup.py#L782

njsmith:
Huh. Weird, but convenient!

PR #14, opened 2016-01-28
Rewrite the section on detecting platform compatibility

njsmith:
After mailing list discussion, get rid of the config file and switch to
"import _manylinux" as the way to configure explicit compatibility
settings.

### DIFF
--- a/pep-513.rst
+++ b/pep-513.rst
@@ -240,7 +240,7 @@ well as the latest releases of Python and ``pip``.
 Auditwheel
 ----------
 
-The second tools is a command line executable called ``auditwheel`` [10]_ that
+The second tool is a command line executable called ``auditwheel`` [10]_ that
 may aid in package maintainers in dealing with third-party external
 dependencies.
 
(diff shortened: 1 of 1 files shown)


rmcgibbo:
Okay. I'm fairly neutral on this part of the PEP (this is mainly a feature for the distros, and I don't really know their perspectives well). But everything in this seems fine to me, and the explanation is quite exhaustive and clear.

It's definitely an improvement over the previous `/etc/python/compatibility.cfg` plan for users that want to install a separate Python installation in their home directory that isn't governed by whatever policy the distro wants to specify for `/usr/bin/python`.

njsmith:
> this is mainly a feature for the distros, and I don't really know their perspectives well

Eh, I guess really this is mostly a feature for us to wave prophylactically at anyone who starts complaining about Linux ABI being sooo hard ("look, you can opt out!"). I guess it's better to have an escape hatch then not, but I'd even be fine just leaving it out myself... I was hoping the distro maintainers would actually weigh in :-/. But, whatever, putting it in is trivial and cheap as well, and it'd be good to get this done and move on to the fun part :-).

(...I guess that means "the fun part" is writing a patch for pip. And waiting for them to make a release with the patch included. But *then* there's the "upload all the wheels!" part which should be more fun.)

pep-513.rst
```diff
@@ -347,31 +347,64 @@ think otherwise.
...
+  the profile changes its ABI in a backwards-incompatible way)
 * A linux distribution that is too old (e.g. RHEL 4)
 * A linux distribution that does not use ``glibc`` (e.g. Alpine Linux, which is
   based on musl ``libc``, or Android)
-* Eventually, in the future, there may exist distributions that break
-  compatibility with this profile
 
-To handle the first two cases, we propose the following simple and reliable
-check: ::
+Therefore, we propose a two-pronged approach. To catch the the first
```
xavfernandez: extra `the`

rmcgibbo:
I think nick endorsed this, so let's merge it.

njsmith:
Yep! I just made a tiny tweak (adding a check to the top of `is_manylinux1_compatible` that we're actually on linux x86-64 / i386), and updated the post date; now I'll merge and re-post.

PR #16, opened 2016-01-29
Fix OpenSSL in Dockerfile

rmcgibbo:


rmcgibbo:
lol @ "Also, should probably use https I guess, even if we are checking the sha256 -- belt and suspenders." in 5e46561

njsmith:
Is it possible to delete the OPENSSL_VERSION from the environment afterwards? Maybe I'm just paranoid, but if it's possible to misuse something (e.g. someone treating this as part of the docker image's public api that is helpfully telling them the latest version of openssl that they should download, so that they don't have to maintain their own security), then IME people will *very* often find a way to misuse it...

And yes, I can't take credit for that saying, but it is one of my favorite security dicta :-)

rmcgibbo:
okay, i dont think it'll be preserved this way

njsmith:
Thanks for putting up with my nitpicking :-)

PR #17, opened 2016-01-30
Refactor Dockerfile into component scripts

matthew-brett:
Refactor some Dockerfile code into a couple of scripts.

Add builds of Python 2.6 and 3.3.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,88 +1,10 @@
 FROM centos:5.11
 MAINTAINER The ManyLinux project
-
-# Python versions installed in /opt/python. Could add more, if desired.
-ENV PY_VERS="2.7.11 3.4.4 3.5.1"
-
-ENV LC_ALL=en_US.UTF-8
-ENV LANG=en_US.UTF-8
-ENV LANGUAGE=en_US.UTF-8
-
-# Dependencies for compiling Python that we want to remove from the final
-# image after compiling Python
-ENV PYTHON_COMPILE_DEPS="zlib-devel bzip2-devel ncurses-devel sqlite-devel readline-devel tk-devel gdbm-devel db4-devel libpcap-devel xz-devel"
-
-# Libraries that are allowed as part of the manylinux1 profile
-ENV MANYLINUX1_DEPS="glibc-devel libstdc++-devel glib2-devel libX11-devel libXext-devel libXrender-devel  mesa-libGL-devel libICE-devel libSM-devel ncurses-devel"
-
-# EPEL support
-ADD https://dl.fedoraproject.org/pub/epel/5/x86_64/epel-release-5-4.noarch.rpm /
-# Dev toolset (for LLVM and other projects requiring C++11 support)
-ADD http://people.centos.org/tru/devtools-2/devtools-2.repo /etc/yum.repos.d/
-RUN rpm -Uvh --replacepkgs epel-release-5*.rpm && rm -f epel-release-5*.rpm
-
-
-# Development tools and libraries
-RUN yum -y install curl wget bzip2 make git patch unzip bison yasm autoconf automake which file \
-                   kernel-devel-`uname -r` \
-                   devtoolset-2-binutils devtoolset-2-gcc \
-                   devtoolset-2-gcc-c++ devtoolset-2-gcc-gfortran \
-                   ${PYTHON_COMPILE_DEPS}
+ENV LC_ALL en_US.UTF-8
+ENV LANG en_US.UTF-8
+ENV LANGUAGE en_US.UTF-8
 ENV PATH /opt/rh/devtoolset-2/root/usr/bin:$PATH
 ENV LD_LIBRARY_PATH /opt/rh/devtoolset-2/root/usr/lib64:/opt/rh/devtoolset-2/root/usr/lib
-
-
-# Compile the latest Python releases.
-# (In order to have a proper SSL module, Python is compiled against openssl-1.0.2e
-# which is linked statically, and then we delete openssl afterwards)
-# Note: 3.2 and earlier need --enable-unicode=ucs4. We don't plan to ever
-# support 3.0-3.2, so really for just this just means Python 2.
-
-RUN export OPENSSL_VERSION=openssl-1.0.2f && \
-    export OPENSSL_SHA256=932b4ee4def2b434f85435d9e3e19ca8ba99ce9a065a61524b429a9d5e9b2e9c && \
-    echo "${OPENSSL_SHA256}  ${OPENSSL_VERSION}.tar.gz" > ${OPENSSL_VERSION}.tar.gz.sha256 && \
-    wget http://www.openssl.org/source/${OPENSSL_VERSION}.tar.gz && \
-    sha256sum -c ${OPENSSL_VERSION}.tar.gz.sha256 && \
-    tar -xzf ${OPENSSL_VERSION}.tar.gz && \
-    (cd ${OPENSSL_VERSION} && ./config no-ssl2 no-shared -fPIC --prefix=/usr/local/ssl && make && make install) && \
-    rm -rf ${OPENSSL_VERSION} ${OPENSSL_VERSION}.tar.gz ${OPENSSL_VERSION}.tar.gz.sha256 && \
-    for PYVER in ${PY_VERS}; do  \
-        if [[ "$(echo $PYVER | cut -d. -f 1)" == "2" ]]; then \
-            UNICODE_OPT="--enable-unicode=ucs4"; \
-        fi; \
-        wget -q https://www.python.org/ftp/python/$PYVER/Python-$PYVER.tgz && \
-        tar -xzf Python-$PYVER.tgz && \
-        (cd Python-$PYVER && \
-            mkdir -p /opt/$PYVER/lib && \
-            LDFLAGS="-Wl,-rpath /opt/$PYVER/lib" ./configure $UNICODE_OPT --prefix=/opt/$PYVER --enable-shared --with-ensurepip=install && \
-            make -j2 && \
-            make install \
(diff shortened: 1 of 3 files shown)


docker/Dockerfile
```diff
@@ -31,48 +30,28 @@ RUN yum -y install curl wget bzip2 make git patch unzip bison yasm autoconf auto
 ENV PATH /opt/rh/devtoolset-2/root/usr/bin:$PATH
 ENV LD_LIBRARY_PATH /opt/rh/devtoolset-2/root/usr/lib64:/opt/rh/devtoolset-2/root/usr/lib
 
+# Shell scripts for build
+RUN git clone https://github.com/matthew-brett/manylinux.git -b refactor-dockerfile
```
njsmith: This can't be right :-)

docker/Dockerfile
```diff
@@ -31,48 +30,28 @@ RUN yum -y install curl wget bzip2 make git patch unzip bison yasm autoconf auto
...
 RUN wget http://nixos.org/releases/patchelf/patchelf-0.8/patchelf-0.8.tar.gz && \
     tar -xzvf patchelf-0.8.tar.gz && \
     (cd patchelf-0.8 && ./configure && make && make install) && \
-    rm -rf patchelf-0.8.tar.gz patchelf-0.8 && \
-    /opt/3.5/bin/pip install git+git://github.com/manylinux/auditwheel.git && \
-    ln -s /opt/3.5/bin/auditwheel /usr/local/bin/auditwheel
+    rm -rf patchelf-0.8.tar.gz patchelf-0.8
 
+RUN /opt/3.5/bin/pip install git+git://github.com/manylinux/auditwheel.git && \
+    ln -s /opt/3.5/bin/auditwheel /usr/local/bin/auditwheel
```
njsmith: Why split this off into its own RUN? (I guess it's good to keep the number of RUN commands down, because docker keeps an intermediate snapshot for each command. They're stored in a compressed way, but still...)

Unrelatedly, @rmcgibbo: maybe we should be switching to the pypi version of auditwheel? Possibly installed as `/opt/3.5/bin/pip install auditwheel==1.0.0`, and then updating that at each release as a cache-buster?

rmcgibbo: Unrelatedly, @rmcgibbo: maybe we should be switching to the pypi version of auditwheel?

Yes, probably. I'll make a pypi release of the latest HEAD.

> Possibly installed as /opt/3.5/bin/pip install auditwheel==1.0.0, and then updating that at each release as a cache-buster?

I'm not really sure what this means.

rmcgibbo:
Probably for "production" it would be good to replace the entire Dockerfile with

```
MAINTAINER <foo>
ENV <bar> <baz> ...
RUN ./all-contents-is-put-in-a-single-shell-script
```

docker/Dockerfile
njsmith: Sorry -- it means that by explicitly mentioning the version in the text of the Dockerfile, we'd prevent docker's caching mechanism for kicking in and saying "hey, I already computed the result of doing `pip install auditwheel`, I don't have to actually run the command or check pypi or anything".

Guess quay.io doesn't use this cache anyway so it probably doesn't matter :-) Need to put a note in the auditwheel release checklist to trigger a manual rebuild on quay.io after each release... (do you need access?)

njsmith:
Yeah, I was just thinking that actually... esp. I think right now because we do `RUN yum install ...`, ..., `RUN yum erase ...`, then we end up making everyone download the intermediate image that has all the packages that we then deleted...

njsmith:
Cf. the "holy build box" dockerfile: https://github.com/phusion/holy-build-box/blob/master/Dockerfile-64

docker/Dockerfile
rmcgibbo: okay, auditwheel==1.1.0 is up on pypi now.

docker/Dockerfile
matthew-brett: An artefact of the possibly silly attempt to put the files in the same repo as the Dockerfile.  Alternative suggestions welcome.   Obviously, barring better suggestions, would in due course be:

```
RUN git clone https://github.com/manylinux/manylinux.git
```

rmcgibbo: I think the dockerfile ADD command works fine with relative paths?

matthew-brett:
One big script would be fine I guess, but it's useful to have bite-sized pieces so it's easier to debug stuff.  I guess this would only be a big script calling some small scripts.

rmcgibbo:
I don't have a significant preference between either of those options.

matthew-brett:
The point about adding the packages and then deleting them seems like a good one.  If y'all agree, I'll make it one big script, calling other scripts, on the holy build box pattern.

docker/Dockerfile
matthew-brett: Ah - I'm new to docker, so I didn't know ADD or COPY - thanks - that's the sensible way.

njsmith:
Sounds good to me!

docker/Dockerfile
njsmith: Docker is rather weird. Apparently the way it works is:
* Everything inside the directory that contains the Dockerfile is (recursively) uploaded to the docker daemon
* Any of those files which get mentioned in an `ADD` / `COPY` command get placed into the container
* All the rest of those files get thrown away

Therefore the rule is: put the Dockerfile in its own directory with just the stuff it needs, and then use `COPY`. (And apparently [`COPY` is preferred over `ADD`](https://docs.docker.com/engine/articles/dockerfile_best-practices/#add-or-copy).)

matthew-brett:
OK - this one is ready to review now.

njsmith:
I think you bumped into my simultaneous changes in #19... probably the unicode stuff is duplicated, but you probably want to pick up the travis.yml changes.

docker/build_scripts/build.sh
```diff
@@ -0,0 +1,64 @@
+#!/bin/bash
+# Python versions installed in /opt/python. Could add more, if desired.
```
njsmith: Comment has drifted away from the thing being commented on.

docker/build_scripts/build.sh
```diff
@@ -0,0 +1,64 @@
...
+
+# Stop at any error
+set -e
```
njsmith: I'd add `-x` to (which causes the shell to print each command before executing it)

docker/build_scripts/build_utils.sh
```diff
@@ -0,0 +1,101 @@
...
+function check_var {
+    if [ -z "$1" ]; then
+        echo "required variable not defined"
+        exit 1
+    fi
+}
+
+
+function pyver_ge {
+    # Echo 1 if first python version is greater or equal to second
```
rmcgibbo: is this really the easiest way to implement this test? uggh bash.

njsmith: Probably not the easiest way, no -- over the versions that we care about, lexicographic order is actually fine... (Guido has said that there will never be a 3.10; it will be called 4.0 instead. 2.7.10 was the first python release to ever have a >1 digit number, and even then it's still restricted to the micro field which we don't care about here).

njsmith: Or even if we don't trust Guido's crystal ball, then this approach would work for us and is much cleaner: http://stackoverflow.com/a/15324047/1925449

njsmith: Meh, I lied -- CentOS 5 is so old that its `sort` doesn't support `-V` :-(

njsmith: I think this should work (sort options set up to numerically compare first 2 "." separated fields; it's a stable sort so if they match the output will be unchanged, making it a `<=` operation):
```
major_minor_version_lte() {
  echo "$1\n$2" | sort -n -t "." -k 1,1 -k 2,2 -s | head -n 1 | grep -q "$1"
}
```

matthew-brett: I don't think this works for me. What output are you expecting from:

```
major_minor_version_lte 3.3 3.5
major_minor_version_lte 3.5 3.3
```

njsmith: Sorry, over-trusting copy from stackoverflow :-) echo doesn't interpret escape sequences, so it should be:
```
major_minor_version_lte() {
  printf "$1\n$2" | sort -n -t "." -k 1,1 -k 2,2 -s | head -n 1 | grep -q "$1"
}
```
which gives
```
[root@84345d3a5606 /]# major_minor_version_lte 3.3 3.5; echo $?
0
[root@84345d3a5606 /]# major_minor_version_lte 3.5 3.3; echo $?
1
```

matthew-brett: Counter-proposal which generalizes to one through three digit comparison - how about:

```
lex_pyver () {
    echo $1 | awk -F "." '{printf "%03d%03d%03d\n", $1, $2, $3}'
}
```

Then:

```
if [ $(lex_pyver 3.2.1) -lt $(lex_pyver 3.3) ]; then echo "yes"; fi
```

njsmith: The only thing this will ever be used for is to detect the strings "2.6" and "2.7", so as long as it isn't actually going to open a rift to the dungeon dimensions then I really don't care :-)

rmcgibbo: In that case why not just use "==2.{6,7}*" for the check?

njsmith: But if we did that, then how would we overengineer it? ;-)

rmcgibbo:
I looked through 8dd3cbd, and it all looks good to me. So +1 to merge whenever.

njsmith:
LGTM too modulo my two nitpicky comments that are still showing up on the PR "Files changed" display

PR #20, opened 2016-01-30
Fix two substantive bugs in the PEP text

njsmith:
- All modern 32-bit linux uses linux_i686 for its tag, not
  linux_i386. My fault for trusting PEP 425. Confirmed by spinning up a
  real CentOS 5 VM in VirtualBox.
- distutils.util.get_platform() uses hyphens; they get normalized to
  underscore later. Fix the is_manylinux1_compatible() code to match.

### DIFF
--- a/pep-513.rst
+++ b/pep-513.rst
@@ -17,7 +17,7 @@ Abstract
 ========
 
 This PEP proposes the creation of a new platform tag for Python package built
-distributions, such as wheels, called ``manylinux1_{x86_64,i386}`` with
+distributions, such as wheels, called ``manylinux1_{x86_64,i686}`` with
 external dependencies limited to a standardized, restricted subset of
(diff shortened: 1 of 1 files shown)


njsmith:
@rmcgibbo: I'd like to upstream these to the official PEP soon because they're embarrassing :-) Since I guess we're both currently looking at github, any comments? :-)

rmcgibbo:
I agree. No substantial comments -- these seem like good (minor) bugfixes.

rmcgibbo:
Just merge than then email the PEP editors -- is that the protocol?

njsmith:
I guess so? There's no real protocol on our end, but I figured it's nice to go through a PR for double-checking and for a diff URL I can send to distutils-sig. And I can send the diff to the pep editors list.

rmcgibbo:
Sounds good.

PR #23, opened 2016-01-31
WIP: Attempt 32-bit image as well

rmcgibbo:
I'm not sure if this works. If it does, it will also require reworking the deploy infrastructure on quay.io.


rmcgibbo:
>The log length has exceeded the limit of 4 MB (this usually means that the test suite is raising the same exception over and over).

> The job has been terminated

.travis.yml
```diff
@@ -4,13 +4,11 @@ services:
...
 
 script:
   >
     set -ex;
-    docker run -t --rm -v $(pwd):/io buildenv sh -c 'set -e; for INTERP in /opt/*/bin/python; do $INTERP /io/manylinux1-check.py; done';
-    PACKAGES=(numpy);
-    for pkg in "${PACKAGES[@]}"; do
-      docker run -t --rm -v `pwd`:/io buildenv /opt/3.5/bin/pip install --no-binary :all: ${pkg} -vv;
-    done
+    docker run -t --rm -v $(pwd):/io manylinux1_x86_64 sh -c 'set -e; for INTERP in /opt/*/bin/python; do $INTERP /io/manylinux1-check.py; done';
```
matthew-brett: Will this pick up the symlinks?  I mean, are we testing /opt/2.7.11 as well as the identical /opt/2.7 ?

rmcgibbo: Yeah, we are.

njsmith: This particular check takes about 0.1 seconds so I figured we might as well just run it on everything.

njsmith:
For docker pushing from travis: this seems to give the basic idea: https://sebest.github.io/post/using-travis-ci-to-build-docker-images/

Give me your quay.io username so I can add you to the group?

rmcgibbo:
`rmcgibbo`

njsmith:
Sent you an invite. And then you should be able to go [here](https://quay.io/organization/pypa?tab=robots) and get the credentials for the "bot account" that we'll use for pushing from travis

rmcgibbo:
Okay, then I'll use `travis encrypt` to encrypt those credentials and add them to the .travis.yml script. That sounds correct?

njsmith:
I think so :-)

On Sun, Jan 31, 2016 at 8:38 PM, Robert T. McGibbon <

> Okay, then I'll use travis encrypt to encrypt those credentials and add
> them to the .travis.yml script. That sounds correct?
>
> —
> <https://github.com/pypa/manylinux/pull/23#issuecomment-177761271>.
>



-- 

docker/build_scripts/build_utils.sh
```diff
@@ -30,9 +30,9 @@ function do_python_build {
     if [ $(lex_pyver $py_ver) -lt $(lex_pyver 3.3) ]; then
         local unicode_flags="--enable-unicode=ucs4"
     fi
-    LDFLAGS="-Wl,-rpath /opt/$py_ver/lib" ./configure --prefix=/opt/$py_ver --enable-shared $unicode_flags
-    make -j2
-    make install
+    CFLAGS="-Wformat" LDFLAGS="-Wl,-rpath /opt/$py_ver/lib" ./configure --prefix=/opt/$py_ver --enable-shared $unicode_flags > /dev/null
```
njsmith: Not sure what the CFLAGS is for... maybe `CFLAGS="-w"`? ([-w:    Inhibit all warning messages. ](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html))

.travis.yml
```diff
@@ -4,13 +4,11 @@ services:
...
+    docker run -t --rm -v $(pwd):/io manylinux1_i686 sh -c 'set -e; for INTERP in /opt/*/bin/python; do $INTERP /io/manylinux1-check.py; done';
```
njsmith: Let's just move these into the end of the build scripts? Really the less we have to do inside travis.yml the better :-)

docker/build_scripts/build_utils.sh
rmcgibbo: I added it in particular for https://bugs.python.org/issue17547 when compiling Python 2.6

rmcgibbo: To silence a bunch of warnings like

```
Include/modsupport.h:27:1: warning: ‘PyArg_ParseTuple’ is an unrecognized format function type [-Wformat=]
 PyAPI_FUNC(int) PyArg_ParseTuple(PyObject *, const char *, ...) Py_FORMAT_PARSETUPLE(PyArg_ParseTuple, 2, 3);
```

njsmith: Ah. Well, I don't understand everything from that link, but I guess a comment wouldn't go amiss :-)

rmcgibbo:
I think I did it correctly, but don't really know if this is going to deploy correctly after it's merged.

njsmith:
Well, let's try it. What's the worst that could happen?

rmcgibbo:
It would also be good to refactor the travis file to use a build matrix so that the two dockerfiles can be built in parallel, perhaps.

njsmith:
yeah, I was thinking about that... the downside is that then I think if one build works and one doesn't, then one will deploy and the other won't, which could be weird or bad?

rmcgibbo:
As long as the whole thing doesn't time out (which it doesn't seem to, but it's pretty close) then it's okay. Otherwise, it'll probably be necessary.

njsmith:
I think the time limit is 50 minutes, so we're only at half that? 25 minutes is annoyingly long though :-/

rmcgibbo:
Oh, I thought it was slower than 25 for some reason, but yeah.

On Sun, Jan 31, 2016 at 10:07 PM, Nathaniel J. Smith <

> I think the time limit is 50 minutes, so we're only at half that? 25
> minutes is annoyingly long though :-/
>
> —
> <https://github.com/pypa/manylinux/pull/23#issuecomment-177786418>.
>



-- 
-Robert

njsmith:
https://docs.travis-ci.com/user/customizing-the-build/#Build-Timeouts

So actually it looks like the limit is 50 minutes per command, so it's separate for the two `docker build` calls.

njsmith:
Well, that didn't work, for reasons that I totally don't understand: https://travis-ci.org/pypa/manylinux/builds/106144601

rmcgibbo:
My one guess was that it the deploy script needs to be executable, d465873805719e67e3869ed914beb03edd537c3e.

njsmith:
Well, it broke differently...

rmcgibbo:
looks like a problem with the environment variables not being exported. i'm testing the syntax in a different repo that builds more quickly.

njsmith:
Travis is so terrible. If only everything else weren't worse.

rmcgibbo:
Okay, the env variables are set now. I just messed it up the first time.

njsmith:
Oh YAML why do you have to be like that.

njsmith:
...so close. apparently we need `push -f`.

docker/Dockerfile-i686
```diff
@@ -0,0 +1,11 @@
+FROM phusion/centos-5-32
+MAINTAINER The ManyLinux project
+
+ENV LC_ALL=en_US.UTF-8
+ENV LANG=en_US.UTF-8
+ENV LANGUAGE en_US.UTF-8
+ENV PATH /opt/rh/devtoolset-2/root/usr/bin:$PATH
+ENV LD_LIBRARY_PATH /opt/rh/devtoolset-2/root/usr/lib64:/opt/rh/devtoolset-2/root/usr/lib
```
matthew-brett: Is the `lib64` correct for the i686 build?

rmcgibbo: Just checked (`docker run --rm -i -t quay.io/pypa/manylinux1_i686 /bin/bash`), and the directory exists, it's just empty.

PR #25, opened 2016-02-01
Rewrite README for the new post-PEP 513 world

njsmith:


njsmith:
@rmcgibbo: review appreciated. Also note that there's a chunk in here that perhaps should point to the new pypa/manylinux-demo (or whatever it ends up being called)

rmcgibbo:
This seems like a good overview to me. With the PEP merged this repo is kind of a grab-bag, and I think the README explains that pretty well. The only minor nitpick is that you rewrote it in markdown, so you should change the file extension.

njsmith:
Hah, thank you. It was actually a bizarre frankenstein markup using ReST + markdown links. I never thought the day would come when I found markdown link syntax intuitive, but here we are... fixed.

There is definitely a question about whether this grab-bag repo is the best way of doing things now -- I guess it might be neater to split off the docker images or something? But OTOH probably we will never touch the PEP text again.

rmcgibbo:
I think it's fine to keep it the PEP here. Once the dust settles, maybe it's something we can revisit it, but I'm not worried about the repo organization now.

PR #31, opened 2016-02-08
PEP updates

njsmith:
- Remove the UCS-4 requirement in favor of pip/wheel's new SOABI
  support.
- Mandate that manylinux1 extensions *not* link against
  libpythonX.Y.so.1, for complicated compatibility-related reasons.
- Update link for the docker images to their new PyPA home.

### DIFF
--- a/pep-513.rst
+++ b/pep-513.rst
@@ -215,17 +215,17 @@ libpythonX.Y.so.1
 -----------------
 
 Note that ``libpythonX.Y.so.1`` is *not* on the list of libraries that
-a ``manylinux1`` extension is allowed to link to. It turns out that
-this is unnecessary in almost all cases: the way ELF linking works,
-extension modules that are loaded into the interpreter automatically
-get access to all of the interpreter's symbols, regardless of whether
-or not the extension itself is explicitly linked against
-libpython. Furthermore, explicit linking to libpython creates problems
-in the common configuration where Python is not built with
-``--enable-shared``. In particular, on Debian and Ubuntu systems,
-``apt install pythonX.Y`` does not even install ``libpythonX.Y.so.1``,
-meaning that any wheel that *did* depend on ``libpythonX.Y.so.1``
-could fail to import.
+a ``manylinux1`` extension is allowed to link to. Explicitly linking
+to ``libpythonX.Y.so.1`` is unnecessary in almost all cases: the way
+ELF linking works, extension modules that are loaded into the
+interpreter automatically get access to all of the interpreter's
+symbols, regardless of whether or not the extension itself is
+explicitly linked against libpython. Furthermore, explicit linking to
+libpython creates problems in the common configuration where Python is
+not built with ``--enable-shared``. In particular, on Debian and
+Ubuntu systems, ``apt install pythonX.Y`` does not even install
+``libpythonX.Y.so.1``, meaning that any wheel that *did* depend on
+``libpythonX.Y.so.1`` could fail to import.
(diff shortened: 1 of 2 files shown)


pep-513.rst
```diff
@@ -216,6 +211,79 @@ the list above to remove libraries that have turned out to be
...
+3.3 and greater use a different storage method that always supports
+UCS-4). If we want to make sure UCS-2 wheels don't get installed into
+UCS-4 CPythons and vice-versa, then something must be done.
+
+An earlier version of this PEP included a requirement that manylinux1
+wheels targeting these older CPython versions should always use the
+UCS-4 ABI. But then, in between the PEP's initial acceptance and its
+implementation, ``pip`` and ``wheel`` gained first-class support for
+tracking and checking this aspect of ABI compatibility for the
+relevant CPython versions, which is a better solution. So we now allow
```
rmcgibbo: I'm not sure this history of the PEP is required.

pep-513.rst
```diff
@@ -216,6 +211,79 @@ the list above to remove libraries that have turned out to be
 problematic or add libraries that have turned out to be safe.
 
 
+libpythonX.Y.so.1
+-----------------
+
+Note that ``libpythonX.Y.so.1`` is *not* on the list of libraries that
+a ``manylinux1`` extension is allowed to link to. It turns out that
```
rmcgibbo: Remove "It turns out that"

rmcgibbo:
I agree with all of the changes. In a couple places (inline comments), I thought that the tone was perhaps a little bit too colloquial, but that's just a minor nitpick.

pep-513.rst
```diff
@@ -216,6 +211,79 @@ the list above to remove libraries that have turned out to be
...
+
+This is a rather complex and subtle issue that extends beyond
+the scope of ``manylinux1``; for more discussion see: [9]_, [10]_,
+[11]_.
+
+UCS-2 vs UCS-4 builds
+---------------------
+
+All versions of CPython 2.x, plus CPython 3.0-3.2, can be built in two
+ABI-incompatible modes: one where Unicode is stored using UCS-2, and
```
matthew-brett: It's actually stored as UTF16 - see https://mail.python.org/pipermail/python-dev/2008-July/080892.html

pep-513.rst
njsmith: PEP's normally spend a lot of time talking about other things that were considered but rejected, etc.; given that we're actually changing the PEP after it was accepted I think it's a good idea to record the reasoning.

pep-513.rst
njsmith: okay

pep-513.rst
njsmith: I'm... just going to ignore this I think :-). (But thank you for the feedback!) The distinction truly doesn't matter for our current purposes, and the configure script and C API both call it UCS-2, so... I tried a few ways of getting this information in there but they were all super clunky.

matthew-brett: I think I would have just done 'a build using the ``--enable-unicode=ucs4`` configure flag stores unicode in UCS-4 / UTF-32 format, whereas a build using ``--enable-unicode=ucs2`` stores unicode in UTF-16. '

njsmith: Ah, that works. Used that, and then tweaked it further :-)

njsmith:
Sounds like there's consensus, and Nick's agreed, so putting this in and sending up to pep-editors

PR #32, opened 2016-02-09
--disable-shared builds in the docker images

rmcgibbo:
This implements one of the changes for the Docker images pursuant to #31, removing `--enable-shared`.


njsmith:
LGTM

rmcgibbo:
We should probably merge #31 first though.

njsmith:
Should we drop the `-rpath` thing too? (Might also solve the problem we were having with it ending up in wheels?)

rmcgibbo:
yeah, probably. it's no longer necessary.

PR #41, opened 2016-03-23
Python != CPython

njsmith:
[this probably doesn't work yet, but I was getting weird "network is unreachable" errors trying to build the image locally -- remember, the great thing about docker build is that it produces identical results no matter where you run it! -- so pushing up to see where travis breaks]

As we are reminded by #38, CPython is not the only implementation of
Python around. This PR doesn't add support for PyPy or anything else,
but it does rearrange things a bit to make room for them.

User visible changes:
- the full-version-number CPython build directories are moved into
  /opt/_internal to make clear that users shouldn't rely on the details
  of their names. Also, they're called e.g. cpython-2.6.9 now instead of
  just 2.6.9.
- The entries in /opt/python/ are now exactly the ABI tags used by each
  version. So we have e.g. /opt/python/cp27mu which generates wheels
  named like *-py27-cp27mu-manylinux1_*.whl. If you want to build
  your wheel against all CPythons, that's /opt/python/cp*/bin/python.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -58,26 +58,33 @@ directory for source code.
 
 The images currently contain:
 
-- CPython 2.6, 2.7, 3.3, 3.4, and 3.5, installed in ``/opt/<version
-  number><soabi flags>``
+- CPython 2.6, 2.7, 3.3, 3.4, and 3.5, installed in
+  ``/opt/python/<abi-tag>``. The directories are named after the ABI
+  tag for the resulting wheels -- e.g. ``/opt/python/cp27mu`` contains
+  a wide-unicode CPython 2.7 build, and can be used to produce wheels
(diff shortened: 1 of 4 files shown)


njsmith:
huh, it worked

ogrisel:
LGTM.

ogrisel:
The README files needs to be updated to take that new folder structure into account.

njsmith:
Updated README, but also changed my mind about what to call the directories :-) See the last commit message for details, and we'll see whether or not the build passes...

ogrisel:
Alright, LGTM too :)

njsmith:
@ogrisel: want to merge, then? You should have the power :-)

ogrisel:
Done!

PR #42, opened 2016-03-25
BF: fix auditwheel install for new Python paths

matthew-brett:
The path to Python 3.5 has changed; adapt auditwheel install to new
path.

### DIFF
--- a/docker/build_scripts/build.sh
+++ b/docker/build_scripts/build.sh
@@ -54,8 +54,9 @@ tar -xzf patchelf-0.8.tar.gz
 (cd patchelf-0.8 && ./configure && make && make install)
 rm -rf patchelf-0.8.tar.gz patchelf-0.8
 
-/opt/3.5m/bin/pip install git+git://github.com/manylinux/auditwheel.git && \
-ln -s /opt/3.5m/bin/auditwheel /usr/local/bin/auditwheel
+PY35_BIN=/opt/python/cp35-cp35m/bin
+$PY35_BIN/pip install auditwheel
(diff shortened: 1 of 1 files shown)


njsmith:
Ugh, how did our tests miss this? Can we add a `
set -e` somewhere?

Also would be good to point the URL at pypa/auditwheel since the manylinux org is deprecated.

matthew-brett:
There's a `set -ex` at the top of the file already, but it looks like the script ran through anyway:

https://travis-ci.org/pypa/manylinux/builds/118358419#L2051

Also, maybe it's time to install auditwheel from pip rather than directly from git master.

njsmith:
Ah-hah, the problem is that `set -e` doesn't apply to `&&` chains. If we delete the `&& \` after the `pip install`, then it should work correctly.

@rmcgibbo: opinion on whether we're ready to take auditwheel from pip? I believe right this moment there are fixes in auditwheel master than aren't released yet?

rmcgibbo:
Pulling auditwheel from pypi is fine (The last release on pypi is 3 days ago).

njsmith:
ah cool

On Fri, Mar 25, 2016 at 9:22 PM, Robert T. McGibbon <

> Pulling auditwheel from pypi is fine (The last release on pypi is 3 days
> ago).
>
> —
> You are receiving this because you commented.
> <https://github.com/pypa/manylinux/pull/42#issuecomment-201702973>
>



-- 

matthew-brett:
Now installing auditwheel from pip

njsmith:
awesome, thanks!

PR #44, opened 2016-04-02
Use patchelf 0.9 + Nathaniel's fixes

ogrisel:
As soon as @matthew-brett's test show that @njsmith's fixes to patchelf are safe, I think we can merge those to the official manylinux docker image.

https://github.com/pypa/auditwheel/pull/25#issuecomment-204806583

### DIFF
--- a/docker/build_scripts/build.sh
+++ b/docker/build_scripts/build.sh
@@ -48,11 +48,11 @@ mkdir -p /opt/python
 build_cpythons $CPYTHON_VERSIONS
 rm -rf /usr/local/ssl
 
-# Install patchelf and auditwheel (latest)
-curl -sLO http://nixos.org/releases/patchelf/patchelf-0.9/patchelf-0.9.tar.gz
-tar -xzf patchelf-0.9.tar.gz
-(cd patchelf-0.9 && ./configure && make && make install)
-rm -rf patchelf-0.9.tar.gz patchelf-0.9
+# Install patchelf and auditwheel (latest with unreleased bug fixes)
+curl -sLO http://nipy.bic.berkeley.edu/manylinux/patchelf-0.9njs2.tar.gz
+tar -xzf patchelf-0.9njs2.tar.gz
+(cd patchelf-0.9njs2 && ./configure && make && make install)
+rm -rf patchelf-0.9njs2.tar.gz patchelf-0.9njs2
 
 PY35_BIN=/opt/python/cp35-cp35m/bin
 $PY35_BIN/pip install auditwheel
--- a/docker/build_scripts/build.sh
+++ b/docker/build_scripts/build.sh
@@ -49,7 +49,7 @@ build_cpythons $CPYTHON_VERSIONS
 rm -rf /usr/local/ssl
 
 # Install patchelf and auditwheel (latest with unreleased bug fixes)
-curl -sLO http://nipy.bic.berkeley.edu/manylinux/patchelf-0.9njs2.tar.gz
+curl -sLO https://nipy.bic.berkeley.edu/manylinux/patchelf-0.9njs2.tar.gz
 tar -xzf patchelf-0.9njs2.tar.gz
 (cd patchelf-0.9njs2 && ./configure && make && make install)
 rm -rf patchelf-0.9njs2.tar.gz patchelf-0.9njs2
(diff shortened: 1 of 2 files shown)


docker/build_scripts/build.sh
```diff
@@ -48,11 +48,11 @@ mkdir -p /opt/python
 build_cpythons $CPYTHON_VERSIONS
 rm -rf /usr/local/ssl
 
-# Install patchelf and auditwheel (latest)
-curl -sLO http://nixos.org/releases/patchelf/patchelf-0.9/patchelf-0.9.tar.gz
-tar -xzf patchelf-0.9.tar.gz
-(cd patchelf-0.9 && ./configure && make && make install)
-rm -rf patchelf-0.9.tar.gz patchelf-0.9
+# Install patchelf and auditwheel (latest with unreleased bug fixes)
+curl -sLO http://nipy.bic.berkeley.edu/manylinux/patchelf-0.9njs2.tar.gz
```
njsmith: now that I think about it, it's kinda not-so-great that we're installing this directly over a plain unencrypted `http` link to a random server. It means that if `nipy.bic.berkeley.edu` gets 0wned or MITMed, then some large proportion of the internet gets 0wned too (basically everyone using pip on linux). We can't help having a somewhat broad trust footprint (github, travis, docker hub for our base image, fedora, python.org, pypi for auditwheel, the TLS certificate system), but we should still try to keep it down if we can :-).

Maybe the best strategy would be to check the hash, like we do with openssl?

matthew-brett: The nipy machine runs https too.​

rmcgibbo: Presumably once this goes into master on patchelf, we can just pull from github?

matthew-brett: Sadly no - the bootstrap autoconf stuff doesn't run on our image, so I had to run this on a more modern machine and then upload the tarball.

rmcgibbo: That's not really a problem. We could just compile and install a newer version of autoconf on the image.

matthew-brett: Sure - or use the patched tarball via https until NixOS does a new release.

njsmith: checking the hash like we do with openssl would be even more secure than https -- in addition to protecting against attacks on the data in transit, it also protects against someone compromising `nipy.bic.berkeley.edu`.

it's always hard to judge how seriously to take these kinds of security risks -- I guess anyone who compromises `nipy.bic.berkeley.edu` can probably add trojans to all kinds of fun stuff that @matthew-brett builds and then distributes all over the place :-). But I'm a little nervous about this docker image in particular, because srsly anyone who can sneak code into here can pull of a pretty apocalyptic security breach -- there are a lot of linux systems in this world that use `pip install`, and probably all of them are going to effectively end up running arbitrary code picked by anyone who compromises this docker image. So taking some simple precautions to reduce our exposure seems like reasonable prudence :-)

ogrisel: Indeed if someone finds a way to rootkit our manylinux1 gcc or patchelf binaries the consequences could be dramatic. I think we need to think about hashing all the file of that image and keep an offline copy somewhere and setup a process to audit periodically the public manylinux1 images from quay.io so as to be able to detect compromised binaries.

PyPI is similarly a potential target and there is a draft PEP to devise a more robust security model for it: https://www.python.org/dev/peps/pep-0480/

ogrisel: Upper in the same script there is another unsafe http URL:

http://people.centos.org/tru/devtools-2/devtools-2.repo

This one does not have an HTTPS counterpart.

ogrisel:
I am working on adding the sha256 checks for all curl'ed files of this script. Expect an update to this PR soon.

ogrisel:
This is looking good. All the sha256 checks passed. Travis is now building all the Pythons.

I also tried locally to change the reference hash values to verify that the build script properly stops in case of bad hash. It works as expected.

ogrisel:
@njsmith merge?

njsmith:
Thanks!

rmcgibbo:
I just cut a release of auditwheel and uploaded to pypi too.

ogrisel:
Thanks @rmcgibbo!

njsmith:
And #47 should finish this off. :fireworks: thanks everyone, this was a lot more complicated than I expected :-)

ogrisel:
> this was a lot more complicated than I expected :-)

But I was amazed by the speed of problem detection, diagnostic and solution of the team. It was a great experience to take part of this. I expect that @matthew-brett will start uploading manylinux1 wheels for the scipy stack in the coming days. I just posted a message on the scikit-learn mailing list pointing theo @matthew-brett's post on the numpy ML to tell people to test the wheels.

matthew-brett:
Yes, I agree - this was a great team effort, enjoyable, informative, productive.  Thanks to y'all for your fine work.

njsmith:
effective teamwork is truly awesome yeah :-)

PR #45, opened 2016-04-03
Add /usr/local/lib{64} to /etc/ld.so.conf by default

### LINKED ISSUE #26: Add standard build directories to LD_LIBRARY_PATH?
Should we add `/usr/local/lib:/usr/local/lib64`to `LD_LIBRARY_PATH` by default?

I notice openblas, for example, builds to `/usr/loca/lib`.

ogrisel:
Fix #26.

### DIFF
--- a/docker/build_scripts/build.sh
+++ b/docker/build_scripts/build.sh
@@ -20,7 +20,7 @@ PYTHON_COMPILE_DEPS="zlib-devel bzip2-devel ncurses-devel sqlite-devel readline-
 MANYLINUX1_DEPS="glibc-devel libstdc++-devel glib2-devel libX11-devel libXext-devel libXrender-devel  mesa-libGL-devel libICE-devel libSM-devel ncurses-devel"
 
 # Get build utilities
-MY_DIR=$(dirname "$BASH_SOURCE[0]}")
+MY_DIR=$(dirname "${BASH_SOURCE[0]}")
 source $MY_DIR/build_utils.sh
 
 # EPEL support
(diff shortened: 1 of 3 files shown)


ogrisel:
You will probably need to run `ldconfig` in your scripts after you install libraries to `/usr/local/lib`.

docker/build_scripts/build.sh
```diff
@@ -23,6 +23,10 @@ MANYLINUX1_DEPS="glibc-devel libstdc++-devel glib2-devel libX11-devel libXext-de
 MY_DIR=$(dirname "$BASH_SOURCE[0]}")
```
piotr-dobrogost: Is left curly bracket missing here?

ogrisel: Indeed. It's weird that bash does not complain in this case. Will push a fix.

njsmith:
Would setting LD_LIBRARY_PATHS in the Dockerfile instead avoid the need for rerunning ldconfig? That seems like the kind of tripwire that will catch a lot of people...

njsmith:
We also might want to canonize Matthew's PKGCONFIG_PATH change: https://github.com/matthew-brett/manylinux-builds/blob/master/common_vars.sh#L16

ogrisel:
@njsmith alright I did the change.

njsmith:
@Matthew-Brett: this should let you clean up those scripts a bit, let us know if it doesn't work for some reason :-)

matthew-brett:
It works, thanks - see : https://github.com/matthew-brett/manylinux-builds/commit/d439645180f5beff446972e22b672b6e6c4a3b65

PR #48, opened 2016-04-07
Add a note about Distrowatch's amazing distro version tables

njsmith:
Just came across these and want to make sure I save the link somewhere
where we can find it when it's time to do manylinux2. Not sure what the
best place is, but this seems like a reasonable option?


ogrisel:
+1. Let's just wait for travis to pass. I saw a low level invalid file data in yum (unrelated to this PR offcourse) but I would like to take the opportunity of that PR to check that the build script still works.

PR #51, opened 2016-04-13
Slim down our Python installs

njsmith:
```
  # BEFORE
  [root@1301fcc7b9d3 /]# du --si -s /opt/_internal/
  1.1G	/opt/_internal/

  # SHRINK IT DOWN
  [root@1301fcc7b9d3 /]# find /opt/_internal -name '*.a' -print0 | xargs -0 rm -f
  [root@1301fcc7b9d3 /]# du --si -s /opt/_internal/
  802M	/opt/_internal/
  [root@1301fcc7b9d3 /]# find /opt/_internal -type f -print0 | xargs -0 -n1 strip --strip-unneeded 2>/dev/null
  [root@1301fcc7b9d3 /]# du --si -s /opt/_internal/
  695M	/opt/_internal/
  [root@1301fcc7b9d3 /]# find /opt/_internal \( -type d -a -name test -o -name tests \) -o \( -type f -a -name '*.pyc' -o -name '*.pyo' \) -print0 | xargs -0 rm -f

  # AFTER
  [root@1301fcc7b9d3 /]# du --si -s /opt/_internal/
  341M	/opt/_internal/
```

So this makes /opt/_internal ~3.2x smaller, and shrinks the overall
(unpacked) filesystem from 1.8 GB to 1.1 GB.

### DIFF
--- a/docker/build_scripts/build.sh
+++ b/docker/build_scripts/build.sh
@@ -74,6 +74,19 @@ yum -y erase wireless-tools gtk2 libX11 hicolor-icon-theme \
 yum -y install ${MANYLINUX1_DEPS}
 yum -y clean all > /dev/null 2>&1
 yum list installed
+# we don't need libpython*.a, and they're many megabytes
+find /opt/_internal -name '*.a' -print0 | xargs -0 rm -f
+# Strip what we can -- and ignore errors, because this just attempts to strip
+# *everything*, including non-ELF files:
+find /opt/_internal -type f -print0 \
+    | xargs -0 -n1 strip --strip-unneeded 2>/dev/null || true
+# We do not need the Python test suites, or indeed the precompiled .pyc and
+# .pyo files. Partially cribbed from:
+#    https://github.com/docker-library/python/blob/master/3.4/slim/Dockerfile
+find /opt/_internal \
+     \( -type d -a -name test -o -name tests \) \
+  -o \( -type f -a -name '*.pyc' -o -name '*.pyo' \) \
+  -print0 | xargs -0 rm -f
 
 for PYTHON in /opt/python/*/bin/python; do
     $PYTHON $MY_DIR/manylinux1-check.py


rmcgibbo:
This seems like a good idea.

ogrisel:
Looks great but it breaks on travis:

```
+ find /opt/_internal -type f -print0
+ xargs -0 -n1 strip --strip-unneeded
The command '/bin/sh -c bash build_scripts/build.sh && rm -r build_scripts' returned a non-zero code: 123
```

Maybe you should just find recursively the `*.so` files in the lib folders and strip all the files in then `bin` folders.

njsmith:
@ogrisel: thanks for the catch -- fixed, I think, let's see what Travis says :-)

I took a slightly different approach of just ignoring errors here -- there are still non-ELF files in bin/ (e.g. pip) so we'd still have to ignore errors unless we got a *lot* more selective, and brute force is easier.

ogrisel:
Travis is happy. Merging!

natefoo:
So uWSGI [directly links against](https://github.com/unbit/uwsgi/blob/master/plugins/python/uwsgiplugin.py#L62) `libpythonX.Y.a`. This may be a pretty obscure case and uWSGI does some nonstandard things, but if you don't link libpython you end up with a bunch of undefined symbols. Is there a good solution for this?

njsmith:
Not sure what to say about the uWSGI case... it basically wants to ship its own copy of the Python interpreter, right? There's no such thing as a uWSGI wheel, this is for building uWSGI-the-application and embedding Python in it? This is a rather special case that technically doesn't have anything to do with manylinux, and it doesn't seem like the solution is for the manylinux docker image to include an extra 300 MB of data in ever download... or OTOH maybe no-one cares about download size and that's exactly the solution, IDK. Maybe open a new issue explaining what you're trying to do in more detail?

natefoo:
Yeah, correct, it delivers itself as a Python package but it's really an application and I believe you're correct about embedding Python. I built local copies of the manylinux1 images that didn't strip the `libpythonX.Y.a`'s and was able to build uWSGI wheels against them, so I think that is a reasonable solution for this obscure case. If it becomes a more frequent issue perhaps a separate set of images (`pypa/manylinux1-libpython_<arch>`?) should be provided.

Issue #53, opened 2016-04-20
system certificate authorities file not loaded by default


henningpeters:
Our wheel building process contacts an https server, but fails with an `[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed` error.

Looking into it a bit more into detail I found that the system has a CA file at `/etc/pki/tls/certs/ca-bundle.crt`, but it's not loaded by default, probably because the Pythons expect it at a different location?

As a workaround I call `export SSL_CERT_FILE=/etc/pki/tls/certs/ca-bundle.crt`, but it would be great to wire it up such that the system CA file is used by default.

njsmith:
Can you give a reproducible example, so that we can test it ourselves? :-)

henningpeters:
Sure:

```
>>> import ssl
>>> ssl.get_default_verify_paths()
DefaultVerifyPaths(cafile=None, capath=None, openssl_cafile_env='SSL_CERT_FILE', openssl_cafile='/usr/local/ssl/ssl/cert.pem', openssl_capath_env='SSL_CERT_DIR', openssl_capath='/usr/local/ssl/ssl/certs')
```

But the paths don't exist:

```
[root@c0bddd8c075f /]# ls -l /usr/local/ssl/ssl/cert.pem
ls: /usr/local/ssl/ssl/cert.pem: No such file or directory
[root@c0bddd8c075f /]# ls -l /usr/local/ssl/ssl/certs
ls: /usr/local/ssl/ssl/certs: No such file or directory
```

Finally, you can see whether the CAs got loaded with:

```
>>> import ssl
>>> ctx = ssl.SSLContext(ssl.PROTOCOL_SSLv23)
>>> ctx.get_ca_certs()
[]
```

Or see the exception with:

```
>>> import urllib
>>> urllib.urlopen('https://google.com')
...
IOError: [Errno socket error] [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:590)
```

njsmith:
It looks like the `get_ca_certs` check is not diagnostic -- on my up-to-date Debian testing I have:
```
>>> import ssl
>>> ctx = ssl.SSLContext(ssl.PROTOCOL_SSLv23)
>>> ctx.get_ca_certs()
[]
```
but cert checking is working:
```
>>> urllib.urlopen("https://google.com")
<addinfourl at 140090398193136 whose fp = <socket._fileobject object at 0x7f6957123bd0>>
[...]

PR #54, opened 2016-04-21
Use the certifi CA certs bundle

njsmith:
Fixes: #53

### DIFF
--- a/docker/build_scripts/build.sh
+++ b/docker/build_scripts/build.sh
@@ -55,6 +55,19 @@ mkdir -p /opt/python
 build_cpythons $CPYTHON_VERSIONS
 rm -rf /usr/local/ssl
 
+PY35_BIN=/opt/python/cp35-cp35m/bin
+
+# Our openssl doesn't know how to find the system CA trust store
+#   (https://github.com/pypa/manylinux/issues/53)
+# And it's not clear how up-to-date that is anyway
+# So let's just use the same one pip and everyone uses
+$PY35_BIN/pip install certifi
+ln -s $($PY35_BIN/python -c 'import certifi; print(certifi.where())') \
+      /opt/_internal/certs.pem
+# If you modify this line you also have to modify the versions in the
+# Dockerfiles:
+export SSL_CERT_FILE=/opt/_internal/certs.pem
+
 # Install patchelf and auditwheel (latest with unreleased bug fixes)
 curl -sLO https://nipy.bic.berkeley.edu/manylinux/patchelf-0.9njs2.tar.gz
 check_sha256sum patchelf-0.9njs2.tar.gz $PATCHELF_HASH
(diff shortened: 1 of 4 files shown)


docker/build_scripts/build.sh
```diff
@@ -89,5 +100,9 @@ find /opt/_internal \
   -print0 | xargs -0 rm -f
 
 for PYTHON in /opt/python/*/bin/python; do
+    # Smoke test to make sure that our Pythons work, and do indeed detect as
+    # being manylinux compatible:
     $PYTHON $MY_DIR/manylinux1-check.py
+    # Make sure that SSL cert checking works
+    $PYTHON $MY_DIR/ssl-check.py
```
ogrisel: Weird: this script seems to be missing from the PR. travis is green but apparently it did not reach this part of the build script, it exited after 26min (instead of more than 30min usually) with:

```
Broadcast message from root@testing-gce-c03d1c2f-f8aa-4f73-8fd1-b3b312ac2218
	(unknown) at 1:24 ...
The system is going down for power off NOW!
```

https://travis-ci.org/pypa/manylinux/builds/124631630#L3928

It's probably a preemptible VM on GCE but I think it's a failure in travis not to report this as a failure.

ogrisel: The shutdown without restart is a bug in travis-ci/travis-ci#4924.

ogrisel: I will restart the build manually to check that ssl-check.py is indeed missing.

ogrisel: A similar shutdown happened again!

ogrisel:
Ok I got travis to fail as expected by restarting it twice:

```
+ ln -s /opt/_internal/cpython-3.5.1 /opt/python/cp35-cp35m
+ rm -f Python-3.5.1.tgz
+ rm get-pip.py
+ rm -rf /usr/local/ssl
+ /pip install certifi
build_scripts/build.sh: line 62: /pip: No such file or directory
The command '/bin/sh -c bash build_scripts/build.sh && rm -r build_scripts' returned a non-zero code: 1
```

njsmith:
hopefully fixed both those issues, let's see what travis says

ogrisel:
There is an import error in the ssl check python script.

njsmith:
Well... it seems to be working now, but travis timed out :-/

njsmith:
It looks like our [last build on master](https://travis-ci.org/pypa/manylinux/builds/125886771) took 45min 45sec, and the limit is 50min, so I don't think this is a new issue with this PR in particular, it's just that adding all those different SOABI variants has pushed us up to the limit of Travis.

I just restarted the build to see what happens, but I guess we urgently have to either split up our build (presumably into one job that builds the x86-64 image and one that builds that i686 image), or else move to a different build service.

I like the property of our current setup that we don't deploy either build unless both have passed... but given that we check that all PRs pass before merging, it probably is okay in practice (since all builds on master should be succeeding). Or alternatively we could switch to something like [wercker](http://wercker.com/) that has support for more complex pipelines.

ogrisel:
> I like the property of our current setup that we don't deploy either build unless both have passed... but given that we check that all PRs pass before merging, it probably is okay in practice (since all builds on master should be succeeding)

+1. I think this is enough as we should not merge PRs as long as they are red.

PR #58, opened 2016-04-28
Add cmake28 package and symlink it to cmake

ibell:
Cmake is a handy tool for development, and would be nice to have in base image

### DIFF
--- a/docker/build_scripts/build.sh
+++ b/docker/build_scripts/build.sh
@@ -20,7 +20,7 @@ PATCHELF_HASH=d9afdff4baeacfbc64861454f368b7f2c15c44d245293f7587bbf726bfe722fb
 PYTHON_COMPILE_DEPS="zlib-devel bzip2-devel ncurses-devel sqlite-devel readline-devel tk-devel gdbm-devel db4-devel libpcap-devel xz-devel"
 
 # Libraries that are allowed as part of the manylinux1 profile
-MANYLINUX1_DEPS="glibc-devel libstdc++-devel glib2-devel libX11-devel libXext-devel libXrender-devel  mesa-libGL-devel libICE-devel libSM-devel ncurses-devel"
+MANYLINUX1_DEPS="glibc-devel libstdc++-devel glib2-devel libX11-devel libXext-devel libXrender-devel  mesa-libGL-devel libICE-devel libSM-devel ncurses-devel cmake28"
 
 # Get build utilities
 MY_DIR=$(dirname "${BASH_SOURCE[0]}")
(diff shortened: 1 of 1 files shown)


docker/build_scripts/build.sh
```diff
@@ -20,7 +20,7 @@ PATCHELF_HASH=d9afdff4baeacfbc64861454f368b7f2c15c44d245293f7587bbf726bfe722fb
 PYTHON_COMPILE_DEPS="zlib-devel bzip2-devel ncurses-devel sqlite-devel readline-devel tk-devel gdbm-devel db4-devel libpcap-devel xz-devel"
 
 # Libraries that are allowed as part of the manylinux1 profile
-MANYLINUX1_DEPS="glibc-devel libstdc++-devel glib2-devel libX11-devel libXext-devel libXrender-devel  mesa-libGL-devel libICE-devel libSM-devel ncurses-devel"
+MANYLINUX1_DEPS="glibc-devel libstdc++-devel glib2-devel libX11-devel libXext-devel libXrender-devel  mesa-libGL-devel libICE-devel libSM-devel ncurses-devel cmake28"
```
matthew-brett: I think cmake should go in the "development tools and libraries" section, not here, because this is the list of libraries required by the manylinux spec.

docker/build_scripts/build.sh
```diff
@@ -72,6 +72,8 @@ yum -y erase wireless-tools gtk2 libX11 hicolor-icon-theme \
     avahi freetype bitstream-vera-fonts \
     ${PYTHON_COMPILE_DEPS}  > /dev/null 2>&1
 yum -y install ${MANYLINUX1_DEPS}
+# Add a soft-link from cmake28 to cmake
+ln -s `which cmake28` /usr/local/bin/cmake
```
matthew-brett: Do you think it is possible that someone wants the system cmake, and will get tripped up by this?  I've no idea myself - I haven't used the system one.

ibell: Perhaps, the system one is pretty ancient though (2.6.4), so I sort of expect that it would be fine.  At the moment I can't even get cmake28 to install (argh).  It was working yesterday.  I don't understand what is going on, something to do with missing libraries.

docker/build_scripts/build.sh
ibell: That's fair.  Sorry that I put it in the wrong place.

ogrisel:
Squashed and merged.

vitaly-krugl:
@ibell and @ogrisel: The title of the PR is "Add cmake28 package **and symlink it to cmake**", however the actual change doesn't do the **symlink it to cmake** part. Thus:

```
docker run -it quay.io/pypa/manylinux1_x86_64 bash
. . .

[root@48d44e269875 /]# cmake
bash: cmake: command not found  
```

PR #72, opened 2016-06-17
add autoconf 2.69 to the image

### LINKED ISSUE #71: autoconf version
The version of autoconf shipping with CentOS 5 is sufficiently out of date to prevent building various projects (among them libffi). It's easy to install it yourself, but since this seems likely to be a common problem for users of the manylinux1 image I'd be happy to send a PR to install the latest autoconf (2.69, released 4 years ago) as part of the build process if people think it belongs here.

(It really just looks like `RUN wget http://ftp.gnu.org/gnu/autoconf/autoconf-latest.tar.gz && tar zxvf autoconf-latest.tar.gz && cd autoconf* && ./configure && make install`)

reaperhulk:
Fixes #71

### DIFF
--- a/docker/build_scripts/build_utils.sh
+++ b/docker/build_scripts/build_utils.sh
@@ -11,6 +11,8 @@ CURL_DOWNLOAD_URL=http://curl.askapache.com/download
 
 GET_PIP_URL=https://bootstrap.pypa.io/get-pip.py
 
+AUTOCONF_DOWNLOAD_URL=http://ftp.gnu.org/gnu/autoconf
+
 
 function check_var {
     if [ -z "$1" ]; then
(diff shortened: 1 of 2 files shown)


docker/build_scripts/build_utils.sh
```diff
@@ -140,3 +142,24 @@ function build_curl {
     (cd ${curl_fname} && do_curl_build)
     rm -rf ${curl_fname} ${curl_fname}.tar.bz2
 }
+
+
+function do_autoconf_build {
```
matthew-brett: Maybe rename as `do_standard_install` or something?

matthew-brett:
Great - thanks for the neat work.

docker/build_scripts/build_utils.sh
reaperhulk: Done!

docker/build_scripts/build_utils.sh
```diff
@@ -140,3 +142,24 @@ function build_curl {
...
+    make install > /dev/null
+}
+
+
+function build_autoconf {
+    local autoconf_fname=$1
+    check_var ${autoconf_fname}
+    local autoconf_sha256=$2
+    check_var ${autoconf_sha256}
+    check_var ${CURL_DOWNLOAD_URL}
```
matthew-brett: Did you mean to check the curl download URL here?

reaperhulk:
One outstanding question -- do we want to remove the `autoconf` from the yum install? Chances are something else depends on it so it'll be implicitly pulled in during dependency resolution thus making the change semi-pointless, but removing it shouldn't hurt.

matthew-brett:
I guess you could move the autoconf build up to before the Python build steps, just after the yum install, and remove the autoconf install, and see what happens.

reaperhulk:
Moving the autoconf to before the yum install shouldn't really change anything since the RPMs will declare dependencies on the autoconf RPM and won't be looking to see if the binary is available or not. I went ahead and looked in the image to see what depends on autoconf:

```
[root@22e5a419f7ea /]# rpm -q --whatrequires autoconf
automake-1.9.6-3.el5
```

So looks like there's no real point to removing it from the deps list, as automake will pull it right back in.

docker/build_scripts/build_utils.sh
reaperhulk: Sure did, fixed.

matthew-brett:
Sure, I was suggesting moving it up _after_ the yum install before the Python build steps.  I think it's neater to remove autoconf from stated dependencies even if in practice it still gets installed.

reaperhulk:
Ah, I understand. I'll push a commit shortly that does that :)

matthew-brett:
Great - thanks - merging.

njsmith:
Unfortunately due to Travis issues, as of right now this change still hasn't landed in the official images -- see #76

Issue #85, opened 2016-12-20
python-config scripts contain flags to link with libpython


rdb:
I understand from #69 that it is recommended that extension modules be compiled with references to the CPython ABI left undefined, and that `libpythonX.Y.so` is therefore intentionally left omitted from the docker image.

However, the `python-config` scripts in the Python installations still contain `-lpythonX.Y` options.  It would be great if this could be removed so that build scripts that use python-config to detect Python installations work correctly.

njsmith:
That sounds correct to me. Any chance you'd be interested in putting
together a patch?

rdb:
I poked around a bit and I think fixing this would involve either shipping our own version of python-config or monkey patching the existing one.

For Python 2.6 through 3.3, this command should work:
`sed -i "s/'-lpython' *+ *pyver\( *+ *sys.abiflags\)\?/''/" path/to/pythonX.Y-config`

3.4 and above use a shell script instead, for which this simpler patch should work:
`sed -i 's/-lpython${VERSION}${ABIFLAGS}//' /opt/python/cp34-cp34m/bin/python3.4m-config`

Does this seem like an acceptable solution or is this too hacky?

njsmith:
It's pretty hacky, but I don't see what else we can really do. IIUC it's a bug in upstream python-config (specifically, python-config and distutils should always do the same thing, and in this case it sounds like they aren't), and maybe we can fix that in future Python releases (and probably we should), but that's a slow process at best, and some of the Pythons we ship are EOL anyway.

mayeut:
This has been fixed in python versions still shipped in supported images.
The reference to `-lpythonX.Y` requires passing `--embed` which is a very special use case (the static libraries are not available by default but can be unpacked for those very special use cases).

Issue #69, opened 2016-06-04
Linking to libpython.so (or not)


c-martinez:
First of all, apologies in advance -- I'm new to building wheels, so this might be a very basic question but...

I'm trying to build a manylinux wheel for [OpenCV](http://opencv.org/), but the OpenCV [build process](http://docs.opencv.org/master/d7/d9f/tutorial_linux_install.html#gsc.tab=0) for python, requires `PYTHON_LIBRARY` to be set to your `libpython<version>.so`.

As far as I understand, it is not recommended to link to libpython when building wheels. But what would be the alternative?

Is it the case that, as long as OpenCV links their library to `libpython`, it will not be possible to build a manylinux wheel for it?

njsmith:
Have you tried pointing it to libpython.a?

c-martinez:
I don't think there is a libpython.a on the manylinux docker images, unfortunately (I'm testing with quay.io/pypa/manylinux1_i686). I did try using `/usr/lib/libpython2.4.so`, but the resulting wheel is broken. Thanks for the suggestion anyway.

matthew-brett:
Ouch.   I think VTK Python has the same problem - it links directly against the Python dynamic library, making it much harder to build wheels - see [this mailing list thread](http://vtk.1045678.n5.nabble.com/Fwd-OSX-build-links-directly-with-Python-binary-avoidable-td5731734.html).

In the longer term, I believe the correct fix is for the OpenCV build process to allow the Python symbols to be undefined at link time - see the thread above.   Do you know any OpenCV folks who could help out with the build?

c-martinez:
Hmmm... I see... No, don't know any OpenCV folk directly, but perhaps they would be happy to see a PR. I'll see what I can do.

Thanks!

njsmith:
Oh right, libpython.a was totally the wrong suggestion, sorry about that :-)

c-martinez:
@njsmith - It was worth trying :-)

blowekamp:
I stumbled on this issue. With regards to VTK there should be a patch for it:
https://gitlab.kitware.com/vtk/vtk/merge_requests/1511

PR #87, opened 2016-12-21
Fix python-config: no longer links with libpython.

rdb:
This monkey patches the python-config script so that build scripts that use it don't try to link to libpythonX.X.so, which is absent from the manylinux distribution.  See #69 for more information.

This also adds `python-config` -> `python3-config` symlink to Python 3 installs, analogous to the existing `python` -> `python3` symlink.

Closes: #85

### DIFF
--- a/docker/build_scripts/build_utils.sh
+++ b/docker/build_scripts/build_utils.sh
@@ -57,6 +57,13 @@ function do_cpython_build {
     # bin/python.
     if [ -e ${prefix}/bin/python3 ]; then
         ln -s python3 ${prefix}/bin/python
+        ln -s python3-config ${prefix}/bin/python-config
+    fi
+    # Remove -lpython from the python-config script.
+    if [ $(lex_pyver $py_ver) -lt $(lex_pyver 3.4) ]; then
+        sed -i "s/'-lpython' *+ *pyver\( *+ *sys.abiflags\)\?/''/" $(readlink -e ${prefix}/bin/python-config)
+    else
+        sed -i 's/-lpython${VERSION}${ABIFLAGS}//' $(readlink -e ${prefix}/bin/python-config)
     fi
     ${prefix}/bin/python get-pip.py
     ${prefix}/bin/pip install wheel


njsmith:
Before we merge this and forget about it, can you add a link to the
bugs.python.org bug report for it?

rdb:
I'm reluctant to do so, because I'm not quite convinced that this is a good change to make to Python in general.  It would break too many existing and valid use cases.

 * I see the use of this approach for building universally deployable wheels, but leaving symbols undefined when linking can be considered sloppy.  Projects may also link with --no-undefined for this reason.  In many situations there is no harm in linking to libpython.
 * Platforms may not necessarily enable the use of undefined symbols or use a flat namespace for symbol lookup by default (eg. Mac).
 * A valid use case for linking to libpython is to embed the interpreter.  These use cases still need a straightforward way to link to the Python library.  This is also relevant when freezing Python applications.
This argument could also apply to manylinux (since people may want to ship programs that embed the interpreter), but manylinux doesn't even have libpython, rendering that use case moot.

njsmith:
The thing about pushing it upstream is that while it's debatable whether Python extensions should link against -lpython, right now upstream is *inconsistent* about it: if you build CPython with `--disable-shared` (like we do) then distutils will *not* use `-lpython` when building extensions, but apparently `python-config` does. That's a bug no matter how you look at it. We didn't invent this thing with not linking to libpython.so, it comes directly from upstream :-). I actually agree that it would make more sense for all extensions to link against libpython.so, but it turns out that a common, standard, upstream-supported configuration on Linux is to ship python *without any libpython.so included*, so...

(And these code paths are all totally platform specific BTW. On Windows you're forced to link against python.dll because there's no such thing as pulling a symbol out of the environment. And on Mac distutils jumps through hoops to disable the default two-level namespace and `--no-undefined` stuff when linking.)

dbogdanov:
What's the status of this fix? Will there be a merge anytime soon? I've stumbled into this issue in my project.

njsmith:
I still think we need to at least file an issue upstream before applying patches to our copy of Python's source code.

dbogdanov:
I've created an [issue on the upstream](https://bugs.python.org/issue33457). You can expand it with more details if necessary.

trishankatdatadog:
Closing because too old. Please reopen if still interested. Thanks!

PR #89, opened 2016-12-30
Add Python 3.6

### LINKED ISSUE #82: Plans for python3.6 image?
I'm curious whether there are any plans to have a Python3.6 manylinux image out soon after the final release next week.

takluyver:
Also update to the latest bugfix release of Python 3.5, 3.4 and 2.7.

Closes #82.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: c
 sudo: required
+dist: trusty
 services:
   - docker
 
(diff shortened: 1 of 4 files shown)


takluyver:
Spotted this warning while compiling 3.5.2:

```
+ ./configure --prefix=/opt/_internal/cpython-3.5.2 --disable-shared
configure: WARNING: linux/random.h: present but cannot be compiled
configure: WARNING: linux/random.h:     check for missing prerequisite headers?
configure: WARNING: linux/random.h: see the Autoconf documentation
configure: WARNING: linux/random.h:     section "Present But Cannot Be Compiled"
configure: WARNING: linux/random.h: proceeding with the compiler's result
configure: WARNING:     ## -------------------------------------- ##
configure: WARNING:     ## Report this to http://bugs.python.org/ ##
configure: WARNING:     ## -------------------------------------- ##
```

takluyver:
And a problem installing pip on 3.6:

```
+ /opt/_internal/cpython-3.6.0/bin/python get-pip.py
Requirement already up-to-date: pip in /opt/_internal/cpython-3.6.0/lib/python3.6/site-packages
Collecting wheel
  Using cached wheel-0.29.0-py2.py3-none-any.whl
Installing collected packages: wheel
Successfully installed wheel-0.29.0
+ /opt/_internal/cpython-3.6.0/bin/pip install wheel
build_scripts/build_utils.sh: line 62: /opt/_internal/cpython-3.6.0/bin/pip: No such file or directory
The command '/bin/sh -c linux32 bash build_scripts/build.sh && rm -r build_scripts' returned a non-zero code: 1
```

ogrisel:
FYI: The bug in configure for Python 3.5.2 and 3.6.0 with respect to `linux/random.h` has been reported here:

https://bugs.python.org/issue27398

takluyver:
Thanks! Sounds like that bit can be safely ignored, then.

ogrisel:
I have also found the following issue when building Python 3.6.0 in the manylinux docker container:

```
*** WARNING: renaming "_sqlite3" since importing it failed: build/lib.linux-x86_64-3.6/_sqlite3.cpython-36m-x86_64-linux-gnu.so: undefined symbol: sqlite3_stmt_readonly
The following modules found by detect_modules() in setup.py, have been
built by the Makefile instead, as configured by the Setup files:
atexit                pwd                   time               

Following modules built successfully but were removed because they could not be imported:
_sqlite3                                                       
```

ogrisel:
The sqlite issue has been reported here:

https://github.com/ghaering/pysqlite/issues/85

ogrisel:
FYI I am fixing the build script for the pip problem (adding a symlink to pip3 when pip is missing but pip is there).

ogrisel:
@takluyver I pushed my changes to your branch. Let's see if travis likes it. If it works I would be in favor of merging this PR and opening a new issue to track the resolution of sqlite3. I think many projects do not need sqlite3 to build their wheels to it would be useful to provide py3.6 support for them.

To fix the actual issue with sqlite3 I am not sure what to do. Any suggestion @ncoghlan @njsmith @rmcgibbo @matthew-brett?

takluyver:
Thanks, that all sounds good to me.

ogrisel:
> The job exceeded the maximum time limit for jobs, and has been terminated.

I think we need to split the build in two independent travis jobs (one for i686 and one for x86_64). The problem is that that would make it possible to deploy a fix for one architecture while the other is failing.

I think this situation is quite rare and we shall not merge PRs with broken builds anyway. Let me try to do the change.

ogrisel:
The build is green and I opened #90 to track the sqlite3 issue. shall we merge?

njsmith:
> I think this situation is quite rare and we shall not merge PRs with broken builds anyway.

Unfortunately, Travis flakiness means that it's not uncommon for a working build to fail for no reason... still, we don't really have any choice right now.

Issue #90, opened 2016-12-30
sqlite3 is broken under Python 3.6


ogrisel:
Here is the relevant snippet from the build log:

```
*** WARNING: renaming "_sqlite3" since importing it failed: build/lib.linux-x86_64-3.6/_sqlite3.cpython-36m-x86_64-linux-gnu.so: undefined symbol: sqlite3_stmt_readonly
The following modules found by detect_modules() in setup.py, have been
built by the Makefile instead, as configured by the Setup files:
atexit                pwd                   time               

Following modules built successfully but were removed because they could not be imported:
_sqlite3                                                       
```

This issue is being discussed here: ghaering/pysqlite#85

The problem is that apparently Python 3.6 requires a more recent version of libsqlite3 that the one available on old centos. We could build or own sqlite3 from source in the docker image but the configure script currently does not make it possible to pass a path to a custom sqlite3 install.

rmcgibbo:
My first thought is that we should just build the most recent libsqlite into the docker image.

rmcgibbo:
And if we just put it at the regular location (/usr/lib), there's no need to pass a custom location to the Python build system.

njsmith:
It's possible that (a) no-one will ever notice b/c people don't usually use sqlite in their build system, (b) by the time they do notice, it will already be fixed upstream. So just ignoring the problem for now also seems like a reasonable option...

ogrisel:
Alright, let close this for now. If a user of the manylinux docker image finds this issue by googling because she / he is impacted by the issue and upstream has not fixed it yet then we can decide to implement https://github.com/pypa/manylinux/issues/90#issuecomment-269822043 (build newest libsqlite3 and install it in /usr/lib) as a stopgap.

Users: please feel free to voice your opinion on this issue by commenting (or opening  new issue and linking to this one).

NathanDunfield:
I just got hit by this.  My Python module uses Sphinx's "autodoc" feature to build its documentation, which means it imports my module, which in turn imports "sqlite3".  This causes my usual build script to fail for Python 3.6 on the manylinux1 container, even though it works fine for Python 2.7, 3.4, and 3.5.   As a stopgap, I can just reuse the documentation that Python 3.5 generates in my Python 3.6 wheel, but a real fix would be great, since I also can't run my doctest suite.

[...]

PR #93, opened 2017-01-23
Use a modern version of SQLite3

NathanDunfield:
This patch has "build.sh" install a recent version of SQLite3 (3.16.2) into /usr/local/lib prior to building all the Pythons.  The motivation is two-fold:

- First, it fixes issue #90, namely that sqlite3 is broken under Python 3.6.

- Second, the version of SQLite3 that comes with Centos5 is *very* old, prior to the signifiant changes of SQLite 3.6.0 which itself was released in mid-2008.  This causes problems when building a Python module that relies on any of these SQLite features, since (a) one can't run the doctests and (b) things like Sphinx's autodoc only work if you can actually import the module.  

I have successfully tested the resulting Docker container by building wheels for [this project](https://bitbucket.org/t3m/snappy).

### DIFF
--- a/docker/build_scripts/build.sh
+++ b/docker/build_scripts/build.sh
@@ -54,6 +54,15 @@ yum -y install bzip2 make git patch unzip bison yasm diffutils \
 build_autoconf $AUTOCONF_ROOT $AUTOCONF_HASH
 autoconf --version
 
+# Install a more recent SQLite3
+curl -sO https://sqlite.org/2017/sqlite-autoconf-3160200.tar.gz
+tar xfz sqlite-autoconf-3160200.tar.gz
+cd sqlite-autoconf-3160200
+./configure
+make install
+cd ..
+rm -rf sqlite-autoconf-3160200*
+
 # Compile the latest Python releases.
 # (In order to have a proper SSL module, Python is compiled
 # against a recent openssl [see env vars above], which is linked


NathanDunfield:
@njsmith, @ogrisel: Any chance of getting this pull request accepted?  Thanks!

ogrisel:
If you build a python module with a recent version of sqlite3 we expect that the module has no build dependency (does not directly link to libsqlite3 via a compiled extension) but instead uses the pure Python API, right? In that case it should note be a problem to use a more recent version in the manylinux docker container.

ogrisel:
Wouldn't it be better to pass "/usr" as the prefix instead of the default (I assume) "/usr/local" prefix in `./configure`?

@njsmith @rmcgibbo any opinion?

NathanDunfield:
@ogrisel: Yes, this is in the context of using the pure Python API only, not directly linking to libsqlite3.  The problem is that the old sqlite3 in Centos5 doesn't accept certain newer commands in Sqlite3's dialect of SQL, but such SQL commands are just sent from Python to Sqlite3 as strings.

NathanDunfield:
@ogrisel: Yes, libsqlite3.so ends up in /usr/local/lib.

NathanDunfield:
@ogrisel: I put the new sqlite3 in "/usr/local/lib" rather than "/usr/lib" for the benefit of any Centos 5 packages that depend on sqlite3.   Also, if I had installed in "/usr/lib" then I'd have clobbered files that should be under the package manager's control.  

Given your setup, it might make the most sense to set the prefix to something in "/opt", say, "/opt/sqlite3" but then of course you'd need to add a flag to each Python configuration script to make it find that version --- with "/usr/local" this is automatic.

ogrisel:
I think using "/usr/local/lib" is good if Python can find there by default.

ogrisel:
@njsmith @rmcgibbo I think we should merge this one. Any reason not to do so?

njsmith:
No objection.

ogrisel:
Ok merged.

NathanDunfield:
@ogrisel, @njsmith:  Many thanks, I really appreciate you incorporating this!

PR #101, opened 2017-04-04
Update README.rst

linshaodongsam:
Original: “The official version of PEP 513 is stored in the PEP repository, but we also have our own copy here. This is where the PEP was originally written, so if for some reason you really want to see the full history of edits it went through, then this is the place to look.”

Revision: “The official version of PEP 513 is stored in the PEP repository, but if you want to see the full history of edits, we have the original copy here.


trishankatdatadog:
Closing because too old. Please reopen if still interested. Thanks!

Issue #100, opened 2017-04-04
Wheel builds broken by CentOS repository going offline


matthew-brett:
CentOS5 just went end-of-life - see #96 .

In consequence (I guess) the repositories that the docker image expect have disappeared, and any manylinux build trying to access these repositories has broken - e.g.:

* https://travis-ci.org/python-pillow/pillow-wheels/jobs/218267819
* https://travis-ci.org/MacPython/matplotlib-wheels/jobs/218370911

It looks like we need to update the docker image to use still-existing repos such as 

http://vault.centos.org/5.11/

See: https://mail.python.org/pipermail/distutils-sig/2017-April/030362.html

matthew-brett:
Hours I think.   I'll look at it later today.

squeaky-pl:
5.11 just started failing several minutes ago for me.

http://centos.mirrors.nublue.co.uk/5.11/os/x86_64/repodata/repomd.xml: [Errno 14] HTTP Error 404: Not Found

larsbutler:
I just ran into this issue as well.

What about moving to a newer version of CentOS, rather than trying to patch things up with 5?

squeaky-pl:
@larsbutler that needs a PEP, new tag and a process.

matthew-brett:
https://github.com/pypa/manylinux/pull/102 may fix this.  Needs merging, and then we need to wait for the docker images to update.  Maybe another 12 hours if all goes well?

njsmith:
Merged, so new docker images should be available in ~30 minutes?

@larsbutler: We'd love to move to a newer release, but it will take a bit of work and we need help :-) If you're interested in helping out, there's a bit of a checklist here: https://mail.python.org/pipermail/distutils-sig/2017-April/030361.html

larsbutler:
@njsmith Awesome, thanks for resolving this so quickly!

After digging into the build scripts further, I realized how non-trivial such an update would be. :) I'm not sure how helpful I can be, but I'll take a look and at least try to educate myself.

matthew-brett:
Fresh numpy wheels builds working now : https://travis-ci.org/MacPython/numpy-wheels/builds/218374431

wiredfool:
Pillow wheels are working now as well: https://travis-ci.org/python-pillow/pillow-wheels/jobs/218570143

PR #102, opened 2017-04-04
BF: use vault repositories for EOL CentOS 5

matthew-brett:
Main repositories for CentOS 5 now offline; point yum at
vault.centos.org (via IP to specialize to a particular server).

Maybe fixes gh-100.


njsmith:
+1 pending tests passing

PR #104, opened 2017-04-06
Reference specific 5.11 release instead of version 5

### LINKED ISSUE #103: Vault repo config is still broken
Due to the EOL-ing of Centos 5, https://github.com/pypa/manylinux/pull/102 recently patched the repo config. However, this still doesn't work:

```
$ docker run -it quay.io/pypa/manylinux1_x86_64 bash
[root@dac4bb86fa5b /]# yum update
http://107.158.252.35/centos/5/os/x86_64/repodata/repomd.xml: [Errno 14] HTTP Error 404: Not Found
Trying other mirror.
Error: Cannot retrieve repository metadata (repomd.xml) for repository: base. Please verify its path and try again
[root@dac4bb86fa5b /]# 
```

There are two issues I can see:

- the URL results in a 404; it looks like some things were shuffled around in the repo
- pointing to a specific IP does not seem very reliable and defeats the purpose of DNS; why not just point to `vault.centos.org`?

Patching coming shortly.

szilveszter:
Also switched the IP address for the DNS name to avoid any future surprises. Fixes #103

Credit for this fix goes out to Ryan Walker (@ryandub) and Lars Butler (@larsbutler).

### DIFF
--- a/docker/build_scripts/build.sh
+++ b/docker/build_scripts/build.sh
@@ -29,10 +29,11 @@ MANYLINUX1_DEPS="glibc-devel libstdc++-devel glib2-devel libX11-devel libXext-de
 # Centos 5 is EOL and is no longer available from the usual mirrors, so switch
 # to http://vault.centos.org
 # From: https://github.com/rust-lang/rust/pull/41045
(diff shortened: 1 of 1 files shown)


matthew-brett:
Great - thanks.

PR #107, opened 2017-04-17
Bundle copy of epel-release-5.4.noarch.rpm

rmcgibbo:


rmcgibbo:
The EPEL RPM for 5.4 is no longer online. This bundles an old copy of the file. Not sure if it'll work or not though,

ofek:
Tests pass!

ogrisel:
Thanks for the fix :)

PR #114, opened 2017-05-23
Update Python 3.6 version to 3.6.1

rdb:
This was released yesterday.

### DIFF
--- a/docker/build_scripts/build.sh
+++ b/docker/build_scripts/build.sh
@@ -5,7 +5,7 @@
 set -ex
 
 # Python versions to be installed in /opt/$VERSION_NO
-CPYTHON_VERSIONS="2.6.9 2.7.13 3.3.6 3.4.6 3.5.3 3.6.0"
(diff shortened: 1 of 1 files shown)


rmcgibbo:
Looks good to me.

njsmith:
Thanks for your contribution! But unfortunately, it turns out that 3.6.1 has a bug that means we probably can't use it :-(

Specifically, 3.6.1 accidentally broke abi in such a way that building on 3.6.0 produces binaries that work everywhere, but building on 3.6.1 produces binaries that (sometimes) don't work on 3.6.0. Since manylinux is all about building highly compatible binaries, this means we can't really use 3.6.1 as shipped.

I think 3.6.2 will fix this, so our options are either to patch out the bug in our copy of 3.6.1, or to stick with 3.6.0 until 3.6.2 is released. The latter is probably simper :-). 3.6.2 should be coming within the next few weeks.

See https://bugs.python.org/issue29943

rdb:
Closing since 3.6.2 should be released by the end of the month.

PR #120, opened 2017-08-22
Specify /bin/bash as the default command

segevfiner:
Saves having to type it in the `docker run` command.

Reference: https://docs.docker.com/engine/reference/builder/#cmd


njsmith:
Seems reasonable. Sorry for the slow response!

PR #124, opened 2017-11-15
Add cmake symlinks

jschueller:
It might save others the time to reinstall from source like I did before inspecting the image more closely.


jschueller:
ping

PR #125, opened 2017-11-15
Update openssl 1.0.2m

jschueller:


njsmith:
I don't understand the travis errors here at all. Does anyone else have any idea what's going on?

rdb:
@njsmith: I ran into this issue some time ago when building manylinux.  Unfortunately the scripts currently suppress any errors from curl with `-s` which makes it hard to debug such failures.  (I would suggest that we pass `-S` alongside `-s` in order not to hide errors.)

The error has something to do with this version of curl not liking the SSL of the PyPA server.  In my own manylinux builds I eventually gave up and just resorted to [packing get-pip with the build scripts](https://github.com/panda3d/buildbot-panda3d/commit/40140e3bc600b9b0e693ec0aa015fca004d9913d) to work around the issue.

jschueller:
Yes, it appears the system curl has a bug. I bundled the get-pip.py if it doesnt succeed.

njsmith:
I added a small cleanup commit; I think this is good to merge once Travis confirms that I didn't screw anything up :-)

matthew-brett:
Cython wheel builds for Python 2.6 are broken, I suspect it's because Python 2.6 is not compatible with the latest pip / `get_pip.py`.  See: https://travis-ci.org/MacPython/cython-wheels/jobs/304800407#L538

njsmith:
I guess it's time to drop 2.6?

matthew-brett:
I thought you'd say that.

njsmith:
Am I wrong? 😀

matthew-brett:
I'll ask the Cython team.  My point in raising it, is so we know that we have therefore dropped support for Python 2.6.

PR #129, opened 2017-12-01
Update centos before doing anything else.

trishankatdatadog:
The centos page on Docker Hub recommends updating packages when using minor tags (e.g., 5.11).

### DIFF
--- a/docker/build_scripts/build.sh
+++ b/docker/build_scripts/build.sh
@@ -42,6 +42,15 @@ sed -i 's/#\(baseurl.*\)mirror.centos.org\/centos\/$releasever/\1vault.centos.or
 MY_DIR=$(dirname "${BASH_SOURCE[0]}")
 source $MY_DIR/build_utils.sh
 
+# https://hub.docker.com/_/centos/
+# "Additionally, images with minor version tags that correspond to install
+# media are also offered. These images DO NOT recieve updates as they are
+# intended to match installation iso contents. If you choose to use these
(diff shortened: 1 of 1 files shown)


njsmith:
This looks sensible. I think we should just call 'yum clean' once at the
end, though. (Are we?)

ogrisel:
>  I think we should just call 'yum clean' once at the end, though.

Docker generates a layer snapshot after each `RUN` statement, hence the need to call clean at the end of each `RUN` line. The final image size is the sum of the size of all the layers generated by the `Dockerfile`.

One way to avoid that would be to use a multi-stage build: https://docs.docker.com/engine/userguide/eng-image/multistage-build/#before-multi-stage-builds

trishankatdatadog:
Thanks @ogrisel and @njsmith!

Dumb question (not an expert here): is there a reason they recommend issuing a `yum clean` right after the initial update? Would it remove obsolete packages from being used later on?

njsmith:
@trishankatdatadog I believe they just want to emphasize that `yum update` and `yum clean` happen inside the same `RUN` line, like @ogrisel was pointing out. Putting them both into the same build script accomplishes the same purpose.

trishankatdatadog:
@njsmith @ogrisel Okay, I see. Would you like me to remove the `yum clean` line?

njsmith:
@trishankatdatadog yes :-)

trishankatdatadog:
Done!

njsmith:
Thanks!

PR #130, opened 2017-12-04
Check hashes of SQLite and patchelf

trishankatdatadog:
Check hashes of SQLite and patchelf. Also update to new versions.


rmcgibbo:
LGTM. +1

PR #131, opened 2017-12-04
Fail on curl error (e.g., during SSL / TLS)

trishankatdatadog:
Slightly stricter checking on curl: fail on error (e.g, during SSL / TLS session. But this is not critical because:
1. We can always check hashes.
2. We should ideally update curl to the latest version as early as possible.


rmcgibbo:
LGTM. +1

njsmith:
This has several conflicts with #130, whose resolution aren't immediately obvious to me.

trishankatdatadog:
@njsmith Let me take a look as soon as I can!

trishankatdatadog:
@njsmith Should be mergeable now, please take a look? Thanks!

njsmith:
@trishankatdatadog Github's still showing the same three conflicts... not sure what's up with that. I've had it happen before that I had to click the "Resolve conflicts" button twice before Github actually resolved the conflicts. 🤷‍

trishankatdatadog:
@njsmith Resolved again :) GitHub finally reports no conflicts now. Waiting to see if tests pass.

njsmith:
Thanks! Sorry about that... no idea why Github does that, but it's pretty annoying.

PR #132, opened 2017-12-04
Verify GPG signatures on Python source tarballs

trishankatdatadog:
Verify GPG signatures on Python source tarballs before building and installing them.

Includes a copy of public keys downloaded from Python.org over SSL/TLS. Please double-check.


trishankatdatadog:
Probably want to put `gpg` in `PYTHON_COMPILE_DEPS`, so that . it's removed at the end of build.

ogrisel:
LGTM. Thanks for working on improving the security of the build pipeline.

docker/build_scripts/build.sh
```diff
@@ -96,6 +97,8 @@ rm -rf sqlite-autoconf-3160200*
 # statically. We delete openssl afterwards.)
 build_openssl $OPENSSL_ROOT $OPENSSL_HASH
 mkdir -p /opt/python
+# Import public keys used to verify these signatures.
+gpg --import $MY_DIR/cpython-pubkeys.txt
```
ogrisel: Actually please add the source URL (where those public keys where downloaded from) to the inline comment.

ogrisel: I believe it's https://www.python.org/static/files/pubkeys.txt

trishankatdatadog:
My pleasure to improve the pipeline security, and done!

ogrisel:
Merged!

PR #135, opened 2017-12-19
Remove '-e' flag from docker login

njsmith:
This has apparently been dropped, and they decided that breaking
people was cool because they printed a warning for a while and who
uses docker in unattended workflows lol:

    https://github.com/moby/moby/issues/23390

So currently our deploys are broken:

    https://travis-ci.org/pypa/manylinux/builds/318363683

This should fix them.


njsmith:
Going to go ahead and merge this myself, since the actual check is what happens once it lands in master and the deploy step runs...

ogrisel:
Thanks!

PR #138, opened 2018-01-02
Update CPython versions to the latest at time of writing

trishankatdatadog:
If nothing breaks, might be a good idea to do this to benefit from any security fixes.

### DIFF
--- a/docker/build_scripts/build_utils.sh
+++ b/docker/build_scripts/build_utils.sh
@@ -49,7 +49,7 @@ function do_cpython_build {
     fi
     local prefix="/opt/_internal/cpython-${py_ver}${dir_suffix}"
     mkdir -p ${prefix}/lib
-    CFLAGS="" ./configure --prefix=${prefix} --disable-shared $unicode_flags > /dev/null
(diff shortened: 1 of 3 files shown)


docker/build_scripts/build_utils.sh
```diff
@@ -49,8 +49,7 @@ function do_cpython_build {
     fi
     local prefix="/opt/_internal/cpython-${py_ver}${dir_suffix}"
     mkdir -p ${prefix}/lib
-    # -Wformat added for https://bugs.python.org/issue17547 on Python 2.6
-    CFLAGS="-Wformat" ./configure --prefix=${prefix} --disable-shared $unicode_flags > /dev/null
+    CFLAGS="" ./configure --prefix=${prefix} --disable-shared $unicode_flags > /dev/null
```
njsmith: If we're not specifying CFLAGS anymore, then there's no need to set it to an empty string; delete it entirely and make the line start with the `./configure` call.

trishankatdatadog:
@njsmith Done, thanks for the feedback!

PR #140, opened 2018-01-03
Enable hash-checking mode for pip install

### LINKED ISSUE #137: Enable hash-checking mode for pip install?
Hi!

**Description of issue or feature request**:

It might be a good idea to [add hashes](https://pip.pypa.io/en/stable/reference/pip_install/#hash-checking-mode) to packages installed via ```pip```, such as ```certifi``` and ```auditwheel```.

Upside: This should allow a build of ```manylinux``` which uses ```pip``` to detect whether or not packages have been modified against expectations.

Downside: This requires manually updating dependencies (but that might be acceptable in this case, as we do pin down the versions and hashes of many other packages).

**Current behavior**:

No hashes are listed.

**Expected behavior**:

Hashes should be listed and checked.

**Miscellaneous notes**:

I'll send a PR when I find some downtime.

trishankatdatadog:
It might be a good idea to [add hashes](https://pip.pypa.io/en/stable/reference/pip_install/#hash-checking-mode) to packages installed via ```pip```, such as ```certifi``` and ```auditwheel```.

Upside: This should allow a build of ```manylinux``` which uses ```pip``` to detect whether or not packages have been modified against expectations.

Downside: This requires manually updating dependencies (but that might be acceptable in this case, as we do pin down the versions and hashes of many other packages).

This fixes #137.

### DIFF
--- a/docker/build_scripts/build.sh
+++ b/docker/build_scripts/build.sh
@@ -116,11 +116,13 @@ build_cpythons $CPYTHON_VERSIONS
 
 PY36_BIN=/opt/python/cp36-cp36m/bin
 
+# Install certifi and auditwheel
+$PY36_BIN/pip install --require-hashes -r $MY_DIR/py36-requirements.txt
+
 # Our openssl doesn't know how to find the system CA trust store
 #   (https://github.com/pypa/manylinux/issues/53)
 # And it's not clear how up-to-date that is anyway
 # So let's just use the same one pip and everyone uses
-$PY36_BIN/pip install certifi
 ln -s $($PY36_BIN/python -c 'import certifi; print(certifi.where())') \
       /opt/_internal/certs.pem
 # If you modify this line you also have to modify the versions in the
@@ -144,8 +146,6 @@ tar -xzf patchelf.tar.gz
 (cd patchelf-$PATCHELF_VERSION && ./bootstrap.sh && ./configure && make && make install)
 rm -rf patchelf.tar.gz patchelf-$PATCHELF_VERSION
 
-# Install latest pypi release of auditwheel
-$PY36_BIN/pip install auditwheel
 ln -s $PY36_BIN/auditwheel /usr/local/bin/auditwheel
 
 # Clean up development headers and other unnecessary stuff for
--- a/docker/build_scripts/build_utils.sh
+++ b/docker/build_scripts/build_utils.sh
@@ -65,7 +65,7 @@ function do_cpython_build {
     fi
     # Since we fall back on a canned copy of get-pip.py, we might not have
     # the latest pip and friends. Upgrade them to make sure.
-    ${prefix}/bin/pip install -U pip wheel setuptools
+    ${prefix}/bin/pip install -U --require-hashes -r ${MY_DIR}/requirements.txt
     local abi_tag=$(${prefix}/bin/python ${MY_DIR}/python-tag-abi-tag.py)
     ln -s ${prefix} /opt/python/${abi_tag}
 }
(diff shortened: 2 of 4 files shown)


njsmith:
I'm nervous about the security/maintainability trade-off here.

The upside of this is that it gives us some protection against a compromise of PyPI (but only if the attackers modify old releases instead of making new ones), or a MITM attack on our connection to PyPI. But the latter shouldn't happen, since by the time we're running pip we're supposed to have working TLS etc. Plus, the main (only?) reason someone would want to compromise the manylinux docker image is so that they could compromise packages when they're built before they're uploaded to PyPI, so if someone's already pwned PyPI then we've already lost.

The downside is that it substantially increases the maintenance burden for the docker image, because now we have to manually update these hashes every time any of these packages make a release, which is something that happens pretty often. In practice this probably means that we just won't update these packages most of the time, which means that we won't be getting any security fixes that they release. This seems like a bigger problem than the one we were originally trying to solve.

What do you think? Is there some threat model I'm not thinking of?

njsmith:
Note that we currently don't have any kind of cryptographic checks on our actual compilers that we're getting from http://people.centos.org/tru/devtools-2/devtools-2.repo – no TLS, no package signatures, nothing. This is probably our weakest link (aside from attacks on developer laptops with commit access). It does look like people.centos.org supports HTTPS now (but the .repo file there doesn't use it); maybe using our own copy of that repo file with a https:// URL swapped in would be a good way to improve the security there?

trishankatdatadog:
On the upside: I don't think you need to compromise all of PyPI. The attacker just needs to, e.g., steal credentials for one of the few (5-6?) PyPI packages used in manylinux. So, in that case, we would be protected.

On the downside: you are right, it does increase the maintenance burden of updating requirements.txt, but only when there is a _security_ update for one of these PyPI packages. I think that has been always been the case anyway for other software included in manylinux (e.g., openssl had to be updated just a weeks ago). To alleviate the burden, there are some automatic tools (e.g., [pyup](https://pyup.io/)) that can be used to monitor whether pinned dependencies now suffer from known vulnerabilities.

To summarize, I agree with you that there is increased maintenance burden, but I think the need to update software on manylinux for security fixes has always been there anyway. For example, even if you don't fix the PyPI package hashes as in here, you will probably need to build and reupload new Docker images anyway when security updates to these packages arise (because the old images would contain the vulnerabilities these updates address).

I hope this makes sense. What do you, and the other maintainers, think?

trishankatdatadog:
Regarding the devtools-2 repo: it looks like those packages have not been updated in three years. It might just be simplest to bundle a static copy of those packages with this GitHub repository, if possible.

njsmith:
> I think that has been always been the case anyway for other software included in manylinux (e.g., openssl had to be updated just a weeks ago).

Indeed, but right now openssl is the main thing that needs to be updated manually; adding another 7 items to that list is a big change, especially since projects like setuptools release much more frequently than openssl does.

Setting up something like pyup or requires.io would indeed make sense to improve the promptness of our rebuilds, independently of this PR. Do they handle updating the hashes?

trishankatdatadog:
I totally understand, @njsmith, and will leave this decision to the maintainers.

Looking at its [config](https://pyup.io/docs/bot/config/), I believe pyup can update requirements.txt files, but this needs testing.

I know TUF is using pyup. @vladimir-v-diaz, sorry to bother you, but would you happen to know if pyup can automatically update your requirements.txt files for you?

vladimir-v-diaz:
@trishankatdatadog 
Yes.  It also [supports pip hashes](https://github.com/pyupio/pyup/commit/3bf5ac9232f82f7e5bf2db0cf243e2f7402643b8).

trishankatdatadog:
Thanks, @vladimir-v-diaz, good to know!

njsmith:
Okay, I just enabled pyup.io, which I *think* should address all my concerns... I guess we can see how it goes.

trishankatdatadog:
Thank you very much, sir! 🙏  @njsmith

PR #144, opened 2018-01-05
Disable Travis "branch builds" for pyup- branches

njsmith:
pyup.io works by creating a branch in the main repo, and then creating
a PR from that branch to master. Travis therefore ends up testing
these commits twice: once because they're a PR, and once because
they're on a branch in the main repo.

This should hopefully disable the branch-based testing, so that we end
up with each PR being tested only once.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,12 @@ dist: trusty
 services:
   - docker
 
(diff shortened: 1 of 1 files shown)


njsmith:
I don't think there's any way to test this other than to merge it and wait, so I guess I'll do that...

PR #147, opened 2018-01-09
Replace broken(?) curl mirror

trishankatdatadog:
I don't know if it's just me, but the previous mirror for downloading the new curl breaks the old curl (error 51: subject name doesn't match host name). This is the easiest, albeit hacky, fix I could find in the meantime. Someone less tired than I am right now should come up with a better fix.

### DIFF
--- a/docker/build_scripts/build_utils.sh
+++ b/docker/build_scripts/build_utils.sh
@@ -7,6 +7,9 @@ PYTHON_DOWNLOAD_URL=https://www.python.org/ftp/python
 # to the ftp mirror:
 OPENSSL_DOWNLOAD_URL=ftp://ftp.openssl.org/source
 # Ditto the curl sources
+# FIXME: This is about the only mirror that supports bootstrapping over a
+# broken version of curl. Unfortunately, we are hardcoding the directory used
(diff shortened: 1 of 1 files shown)


njsmith:
Heh, I just came up with exactly the same fix: #150 

More context: https://github.com/pypa/manylinux/pull/149#issuecomment-358854620

njsmith:
The *correct* solution is to kill CentOS 5 and hide the body. Hopefully this will last until someone manages that :-)

trishankatdatadog:
👍

trishankatdatadog:
i hate to be the bad news bear, but: https://news.ycombinator.com/item?id=16278620

njsmith:
@markrwilliams so about that PEP...

trishankatdatadog:
cc @ofek

markrwilliams:
@njsmith Fastmail says I sent it to distutils-sig yesterday at 4 PM PDT!  Has it not appeared on the list yet?

markrwilliams:
The PEP has arrived! https://mail.python.org/pipermail/distutils-sig/2018-February/031944.html

PR #150, opened 2018-01-19
Use a different curl URL

njsmith:
So far this particular download URL doesn't require modern SSL
support. Let's see how long that lasts.

CentOS 5 needs to die so bad.


njsmith:
Oh, this has a small mistake and is a dupe of #147 anyway

PR #159, opened 2018-02-23
build git ourselves for TLS 1.2 support

reaperhulk:
As of today GitHub no longer allows connections < TLS 1.2. This breaks git in the manylinux1 containers since they're linked against system OpenSSL. It also breaks building the containers at all because they try to fetch the newer curl from github via the old system curl.

To resolve this we can link our own copy of git against the OpenSSL we link the Pythons against, but there are a few knock-on effects...

1) We have to fetch a curl from a debian mirror (or switch to using Docker's `ADD`) to solve the bootstrap problem. To make this minimally invasive I've just swapped to the debian mirror and downgraded curl for now.
2) We have to reorder our various package compiles to prioritize getting a functional curl ASAP.
3) libcurl (which we need for git) can't be deleted immediately. In fact, I can't get git to build against a static libcurl (it complains about symbols that should be only in libcrypto/libssl), so for the moment I've converted it to a dynamic lib and we never delete it. That *shouldn't* be a big deal as the libcurl we install has a different soversion.
4) We need expat-devel because git depends on libexpat.
5) cmake28 uses system curl still, so that's likely to cause anybody using cmake to fetch remote resources pain in the near future.


njsmith:
fine

reaperhulk:
(This isn't ready for merge yet as git is not actually building properly because it's awful)

reaperhulk:
Looks like it's all building successfully now. This should be ready for review.

docker/build_scripts/build.sh
```diff
@@ -19,9 +19,8 @@ DEVTOOLS_HASH=a8ebeb4bed624700f727179e6ef771dafe47651131a00a78b342251415646acc
 # https://github.com/NixOS/patchelf/commit/2a9cefd7d637d160d12dc7946393778fa8abbc58
 PATCHELF_VERSION=2a9cefd7d637d160d12dc7946393778fa8abbc58
 PATCHELF_HASH=12da4727f09be42ae0b54878e1b8e86d85cb7a5b595731cdc1a0a170c4873c6d
-CURL_ROOT=curl-7.57.0
-# https://github.com/Homebrew/homebrew-core/blob/e3a8622111ecefe444194cade5cca3c69165e26c/Formula/curl.rb#L6
-CURL_HASH=c92fe31a348eae079121b73884065e600c533493eb50f1f6cee9c48a3f454826
+CURL_ROOT=curl_7.52.1
```
xhochy: The reason for the version downgrade is probably due to the mirror switch to Debian?

njsmith:
Well, we're clearly trying to bail out a boat while it sinks, so I guess we better hurry up with the bailing...

(PEP 571 will be what we evacuate to.)

njsmith:
Github's just giving me a "server error" when I try to click merge...

docker/build_scripts/build.sh
reaperhulk: For others who might find this PR, yes this was the reason. I'd prefer to just go to 7.58.0, but that would require switching to `ADD` in docker. I do think we should switch to that and use less bash to build this thing now that `--squash` is an option, but that's best done in `manylinux2`.

PR #162, opened 2018-03-03
manylinux1_x86_64: Try to not install i686/i386

jcfr:


njsmith:
What does this do? Why did you need it?

jcfr:
It decreases the size of the x86_64 image by not installing the i686/i386 packages.

Since I re-used the script of manylinux in a different context, I thought it would make sense to contribute this optimization.

jcfr:
@njsmith  @matthew-brett  Anything I could do move this forward ? Thanks for your review and feedback.

njsmith:
We can't really merge anything until openssl is fixed – right now the build would just fail anyway if we did.

njsmith:
Travis is doing something weird here, going to try to close/reopen and see if that fixes it.

PR #170, opened 2018-03-28
docker/build_script: Update openssl to latest version: 1.0.2n -> 1.0.2o

jcfr:
See https://www.openssl.org/news/secadv/20180327.txt (thanks @natefoo for the link)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ matrix:
 
 script:
   # Workaround Travis issue #9391 (see build_scripts/build_utils.sh - function build_openssl)
-  - (cd docker/build_scripts/ && curl -fsSLO https://www.openssl.org/source/openssl-1.0.2o.tar.gz)
+  - curl -fsS -o docker/openssl-1.0.2o.tar.gz https://www.openssl.org/source/openssl-1.0.2o.tar.gz
   - docker build --rm -t quay.io/pypa/manylinux1_$PLATFORM:$TRAVIS_COMMIT -f docker/Dockerfile-$PLATFORM docker/
 
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,8 @@ matrix:
     - env: PLATFORM="x86_64"
 
 script:
+  # Workaround Travis issue #9391 (see build_scripts/build_utils.sh - function build_openssl)
+  - (cd docker/build_scripts/ && curl -fsSLO https://www.openssl.org/source/openssl-1.0.2o.tar.gz)
   - docker build --rm -t quay.io/pypa/manylinux1_$PLATFORM:$TRAVIS_COMMIT -f docker/Dockerfile-$PLATFORM docker/
 
 
--- a/docker/build_scripts/build.sh
+++ b/docker/build_scripts/build.sh
@@ -9,10 +9,10 @@ CPYTHON_VERSIONS="2.7.14 3.3.7 3.4.7 3.5.4 3.6.4"
 
 # openssl version to build, with expected sha256 hash of .tar.gz
 # archive
-OPENSSL_ROOT=openssl-1.0.2n
-# Hash from https://www.openssl.org/source/openssl-1.0.2n.tar.gz.sha256
-# Matches hash at https://github.com/Homebrew/homebrew-core/blob/99b8ea3594d1f1f78b0fff1fd8ca7d782aa07e13/Formula/openssl.rb#L11
-OPENSSL_HASH=370babb75f278c39e0c50e8c4e7493bc0f18db6867478341a832a982fd15a8fe
+OPENSSL_ROOT=openssl-1.0.2o
+# Hash from https://www.openssl.org/source/openssl-1.0.2o.tar.gz.sha256
+# Matches hash at https://github.com/Homebrew/homebrew-core/blob/1766321103d9780f6e38d3ac7681b8fa42cdca86/Formula/openssl.rb#L11
+OPENSSL_HASH=ec3f5c9714ba0fd45cb4e087301eb1336c317e0d20b575a125050470e8089e4d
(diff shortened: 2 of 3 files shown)


jcfr:
Cc: @matthew-brett @njsmith

jcfr:
Current download failure is quite strange:

```
+ wget ftp://ftp.openssl.org/source/openssl-1.0.2o.tar.gz
--2018-03-28 16:38:40--  ftp://ftp.openssl.org/source/openssl-1.0.2o.tar.gz
           => `openssl-1.0.2o.tar.gz'
Resolving ftp.openssl.org... 194.97.150.234, 2001:608:c00:180::1:ea
Connecting to ftp.openssl.org|194.97.150.234|:21... connected.
Logging in as anonymous ... Logged in!
==> SYST ... done.    ==> PWD ... done.
==> TYPE I ... done.  ==> CWD /source ... done.
==> SIZE openssl-1.0.2o.tar.gz ... 5329472
==> PASV ... couldn't connect to 194.97.150.234 port 16022: No route to host
The command '/bin/sh -c linux32 bash build_scripts/build.sh && rm -r build_scripts' returned a non-zero code: 1
```

Locally, it works:

```
wget ftp://194.97.150.234/source/openssl-1.0.2o.tar.gz
--2018-03-28 16:50:13--  ftp://194.97.150.234/source/openssl-1.0.2o.tar.gz
           => ‘openssl-1.0.2o.tar.gz’
Connecting to 194.97.150.234:21... connected.
Logging in as anonymous ... Logged in!
==> SYST ... done.    ==> PWD ... done.
==> TYPE I ... done.  ==> CWD (1) /source ... done.
==> SIZE openssl-1.0.2o.tar.gz ... 5329472
==> PASV ... done.    ==> RETR openssl-1.0.2o.tar.gz ... done.
Length: 5329472 (5.1M) (unauthoritative)

openssl-1.0.2o.tar.gz                                     100%[=====================================================================================================================================>]   5.08M  1.26MB/s   in 6.4s   

2018-03-28 16:50:21 (816 KB/s) - ‘openssl-1.0.2o.tar.gz’ saved [5329472]

```

jcfr:
Unless someone know why it is failing, next step would be to try locally downloading the travis docker image:

```
travis-job-3078caab-9b1a-4d9c-8ad1-ba3f4e78fc26 travis-ci-garnet-trusty-1512502259-986baf0
```

References:
* https://stackoverflow.com/questions/21053657/how-to-run-travis-ci-locally/49019950#49019950

natefoo:
It appears to be a passive FTP problem, I'm not sure why this wouldn't have turned up before unless Travis has changed the way they handle outgoing ports. It'd probably be fine to switch to https, so [this](https://github.com/pypa/manylinux/blob/e662029e490b9b46df9d1ec0c09e3471e14a8c1b/docker/build_scripts/build_utils.sh#L8):

```sh
OPENSSL_DOWNLOAD_URL=ftp://ftp.openssl.org/source
```

becomes this:

```sh
OPENSSL_DOWNLOAD_URL=https://www.openssl.org/source
```

jcfr:
> It'd probably be fine to switch to https

`ftp` is used because the centos based container do not support https. It is a chicken and egg problem.

A possible solution would be explicitly download the archive before [calling docker](https://github.com/pypa/manylinux/blob/e662029e490b9b46df9d1ec0c09e3471e14a8c1b/.travis.yml#L25-L26) and update the dockerbuild to copy the archive

jcfr:
This seems to be a recent issue now documented in [travis-ci/travis-ci#9391: FTP connections on GCE (a.k.a. `sudo`-required) infrastructure are unreliable](https://github.com/travis-ci/travis-ci/issues/9391)

Look like the way forward is to retry the download.

natefoo:
> I think `ftp` is used because the centos based container do not support https.

Sure enough, it even explains that in the comment right above the line I linked to.

I'll have a look at Travis and see what's changed.

natefoo:
Beat me to it, ha! Thanks.

jcfr:
In the mean time, this last commit should workaround the unreliable ftp download.

jcfr:
@natefoo I also added you as co-author in the last commit. Let me know if you prefer not.

jcfr:
Humm .. no luck with the re-try approach

natefoo:
Rather than subshells, which break `set -e`, you can use a loop like:

```sh
for i in {0..4}; do
    wget ... && break
done || false
```

But, it looks like the "retry" method isn't working anyway.

jcfr:
Thanks. I will remember this. 

I am backtracking these changes and downloading the file before calling `docker --build ...`

natefoo:
Thanks for working on this. I made a small change to reduce some of the file handling complexity in jcfr/manylinux#1.

jcfr:
Nice. The idea by not modifying the Dockerfile was to avoid hardcoding the version in a third place.

njsmith:
The download-before-docker-build trick is a good idea! Maybe we should just do that in general for the bootstrap files (openssl and curl sources) instead of worrying about ftp and carefully chosen mirrors and stuff.

Would you be up for making the same change for curl? And you can drop the comments about working around travis's FTP bug, it's a silly bug but this way is better than what we were doing anyway.

natefoo:
@njsmith Ok, in that case, I'll put in a little effort to adding a sources dir to the context and updating the build utils to be aware of that directory so we don't have to do a bunch of `COPY`s and then moving sources to cwd in the build script.

jcfr:
> Can you squash the commits once it's ready to go?

All set. See https://github.com/pypa/manylinux/pull/172

But for some reason the CI didn't kick in.

> The download-before-docker-build trick is a good idea! Maybe we should just do that in general for the bootstrap files

:+1:

jcfr:
> I'll put in a little effort to adding a sources dir

@natefoo I will let you takes the lead now.

jcfr:
Closing. This is superseded by #172

PR #172, opened 2018-03-28
Update OpenSSL from 1.0.2n to 1.0.2o (squashed)

jcfr:

PR #174, opened 2018-03-29
Prefetch source tarballs for OpenSSL and curl outside the docker build context in Travis

natefoo:
Also, upgrade OpenSSL to 1.0.2o.

This should avoid problems with using the old tools (e.g. wget) in EL5 to fetch these sources. Since people may do Docker Hub, quay.io, etc. automated builds of their own forks from the Dockerfile/context only, prefetching is not required and if sources are not prefetched we fall back to in-container fetching.

There's a lot more work that can be done to reduce duplication in the build scripts, especially with bash indirect references.

Incorporates/supercedes #172, xref #170, thanks @jcfr

### DIFF
--- a/docker/build_scripts/build_utils.sh
+++ b/docker/build_scripts/build_utils.sh
@@ -1,23 +1,6 @@
 #!/bin/bash
 # Helper utilities for build
 
-PYTHON_DOWNLOAD_URL=https://www.python.org/ftp/python
-# XXX: the official https server at www.openssl.org cannot be reached
-# with the old versions of openssl and curl in Centos 5.11 hence the fallback
-# to the ftp mirror:
-OPENSSL_DOWNLOAD_URL=ftp://ftp.openssl.org/source
-# We had to switch to a debian mirror because we can't use TLS until we
-# bootstrap it with this curl + openssl
-CURL_DOWNLOAD_URL=http://deb.debian.org/debian/pool/main/c/curl
-
-GET_PIP_URL=https://bootstrap.pypa.io/get-pip.py
-
-AUTOCONF_DOWNLOAD_URL=http://ftp.gnu.org/gnu/autoconf
-AUTOMAKE_DOWNLOAD_URL=http://ftp.gnu.org/gnu/automake
-LIBTOOL_DOWNLOAD_URL=http://ftp.gnu.org/gnu/libtool
-
-GIT_DOWNLOAD_URL=https://github.com/git/git/archive
-
 
 function check_var {
     if [ -z "$1" ]; then
@@ -117,6 +100,32 @@ function do_openssl_build {
 }
 
 
+function _fetch_source {
+    local with=$1
+    check_var ${with}
+    local file=$2
+    check_var ${file}
+    local url=$3
+    check_var ${url}
+    local dir=$4/  # so you can get dir=/ by not passing a 4th arg
+    [ -d $dir ] || mkdir $dir
+    if [ -f ${dir}${file} ]; then
+        echo "${dir}${file} exists, skipping fetch"
+    elif [ "$with" = "wget" ]; then
+        wget -q -O ${dir}${file} ${url}/${file}
+    else
+        curl -fsSL -o ${dir}${file} ${url}/${file}
+    fi
+}
+
+
+function fetch_source {
+    # This is called outside the build context (e.g. in Travis) to prefetch
+    # source tarballs, where curl exists (and works)
+    _fetch_source curl "$@"
+}
+
+
 function check_sha256sum {
     local fname=$1
     check_var ${fname}
@@ -135,13 +144,8 @@ function build_openssl {
     local openssl_sha256=$2
     check_var ${openssl_sha256}
     check_var ${OPENSSL_DOWNLOAD_URL}
-    #
-    # XXX Workaround Travis issue #9391 (FTP connections on GCE (a.k.a. `sudo`-required) infrastructure are unreliable)
-    #     by downloading the archive outside of the container using https. See .travis.yml
-    #
-    [ -f ${MY_DIR}/${openssl_fname}.tar.gz ] && cp ${MY_DIR}/${openssl_fname}.tar.gz .
     # Can't use curl here because we don't have it yet
-    [ -f ${openssl_fname}.tar.gz ] || wget -q ${OPENSSL_DOWNLOAD_URL}/${openssl_fname}.tar.gz
+    _fetch_source wget ${openssl_fname}.tar.gz ${OPENSSL_DOWNLOAD_URL}
     check_sha256sum ${openssl_fname}.tar.gz ${openssl_sha256}
     tar -xzf ${openssl_fname}.tar.gz
     (cd ${openssl_fname} && do_openssl_build)
@@ -155,7 +159,7 @@ function build_git {
     local git_sha256=$2
     check_var ${git_sha256}
     check_var ${GIT_DOWNLOAD_URL}
-    curl -sSLO ${GIT_DOWNLOAD_URL}/v${git_fname}.tar.gz
+    fetch_source v${git_fname}.tar.gz ${GIT_DOWNLOAD_URL}
     check_sha256sum v${git_fname}.tar.gz ${git_sha256}
     tar -xzf v${git_fname}.tar.gz
     (cd git-${git_fname} && make install prefix=/usr/local LDFLAGS="-L/usr/local/ssl/lib -ldl" CFLAGS="-I/usr/local/ssl/include" > /dev/null)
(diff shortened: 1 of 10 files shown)


docker/build_scripts/build_utils.sh
```diff
@@ -155,7 +159,7 @@ function build_git {
     local git_sha256=$2
     check_var ${git_sha256}
     check_var ${GIT_DOWNLOAD_URL}
-    curl -sSLO ${GIT_DOWNLOAD_URL}/v${git_fname}.tar.gz
+    fetch_source v${git_fname}.tar.gz ${GIT_DOWNLOAD_URL}
```
jcfr: look like `curl` need to be specified

docker/build_scripts/build_utils.sh
```diff
@@ -200,7 +204,7 @@ function build_autoconf {
     local autoconf_sha256=$2
     check_var ${autoconf_sha256}
     check_var ${AUTOCONF_DOWNLOAD_URL}
-    curl -sSLO ${AUTOCONF_DOWNLOAD_URL}/${autoconf_fname}.tar.gz
+    fetch_source ${autoconf_fname}.tar.gz ${AUTOCONF_DOWNLOAD_URL}
```
jcfr: look like `curl` need to be specified

docker/build_scripts/build_utils.sh
```diff
@@ -214,7 +218,7 @@ function build_automake {
     local automake_sha256=$2
     check_var ${automake_sha256}
     check_var ${AUTOMAKE_DOWNLOAD_URL}
-    curl -sSLO ${AUTOMAKE_DOWNLOAD_URL}/${automake_fname}.tar.gz
+    fetch_source ${automake_fname}.tar.gz ${AUTOMAKE_DOWNLOAD_URL}
```
jcfr: look like `curl` need to be specified

docker/build_scripts/build_utils.sh
```diff
@@ -228,7 +232,7 @@ function build_libtool {
     local libtool_sha256=$2
     check_var ${libtool_sha256}
     check_var ${LIBTOOL_DOWNLOAD_URL}
-    curl -sSLO ${LIBTOOL_DOWNLOAD_URL}/${libtool_fname}.tar.gz
+    fetch_source ${libtool_fname}.tar.gz ${LIBTOOL_DOWNLOAD_URL}
```
jcfr: here too

docker/build_scripts/prefetch.sh
```diff
@@ -0,0 +1,25 @@
...
+
+[ -d "$dir" ] || mkdir "$dir"
+
+for name in "$@"; do
+    name=$(echo $name | tr '[:lower:]' '[:upper:]')
+    root=${name}_ROOT
+    ext=${name}_EXTENSION
+    url=${name}_DOWNLOAD_URL
+    file=${!root}${!ext:-.tar.gz}
+    fetch_source $file ${!url} $dir
```
jcfr: Same  thing here, look like the `with` argument is not specified

docker/build_scripts/build_utils.sh
natefoo: There are 2 functions, this calls `fetch_source`, which just calls `_fetch_source curl "$@"`. The idea is that fetching with wget is not something that should be done unless explicitly requested, the default is curl.

docker/build_scripts/build_utils.sh
```diff
@@ -117,6 +100,32 @@ function do_openssl_build {
 }
 
 
+function _fetch_source {
```
jcfr: May be worth documenting the method. Also what about the following signature:

```
_fetch_source <url> <dest> [--method (curl|get)]
```

The `file` would be obtained from the url using `file=$(basename ${url})`

docker/build_scripts/build_utils.sh
jcfr: I just realized, there is `_fetch_source` and `fetch_source`

jcfr: Gotcha. If CI is green that works for me.

docker/build_scripts/build_utils.sh
natefoo: Yeah, I might make another pass at this to pretty things up and reduce duplication after we get the immediate issue fixed.

jcfr:
@njsmith  Would be awesome if you could review and integrate :+1:

jcfr:
Nitpick: @natefoo  May you could squash `Fix curl filename` with its parent commit ?

njsmith:
I don't understand why we have this dance with both wget and curl -- isn't the point of prefetching that we can do it on the host system, which presumably has a working curl? What problem do we solve by using wget sometimes?

natefoo:
We aren't using wget if openssl and curl are prefetched. I left the build context fetch with wget functionality in there because forks won't necessarily build with Travis.

Case in point, the whole reason I discovered the OpenSSL version change yesterday was because I was setting up Hub/quay automated/push builds of a fork I need that will retain libpython*.a. If we got rid of the wget stuff, there'd be no way for those types of builds to work. Plus, right now, you can build the images locally entirely with `docker build`, which also wouldn't be possible if we *required* prefetching (people would have to be aware of and run the prefetch step).

njsmith:
That doesn't answer the question though. We're not doing this prefetching dance because we want to; we're doing it because the vanilla centos5 image *cannot* connect to the hosts that distribute the openssl and curl sources.

Or put another way, if the wget thing actually works, then we shouldn't do prefetching either.

jcfr:
> I discovered the OpenSSL version change yesterday was because I was setting up Hub/quay automated/push builds of a fork

Makes sense. I also discovered the problem because I used derived image.

> if the wget thing actually works, then we shouldn't do prefetching either.

It works without problem on CircleCI. May be we could switch to that ?

jcfr:
In the short term, I think it makes sense to address the issue. Then, we could discuss the best strategy to re-factor the prefetch system. 

Would it be a good compromise to get a branch based of https://github.com/pypa/manylinux/pull/174/commits/2e37903b38eaffcf374fae9a22cb3a445c531b40 (Prefetch source tarballs for OpenSSL and curl outside the docker build context in Travis ) integrated now ?

jcfr:
> Would it be a good compromise to get a branch based of 2e37903 (Prefetch source tarballs for OpenSSL and curl outside the docker build context in Travis ) integrated now ?

For convenience, I re-open the associated PR. https://github.com/pypa/manylinux/pull/172

Thanks @njsmith

njsmith:
> It works without problem on CircleCI. May be we could switch to that ?

Using circleci might work around travis's broken network and  allow the `ftp://...` hack to work, but the only reason we're using ftp is as a gross/fragile/incomplete hack. (In particular, we need both openssl and curl to bootstrap, and the curl project doesn't maintain an ftp server.)

natefoo:
> we're doing it because the vanilla centos5 image *cannot* connect to the hosts that distribute the openssl and curl sources.

Only on Travis, right now, due to the travis-ci/travis-ci#9391. It works fine locally and on other build platforms.

jcfr:
> In particular, we need both openssl and curl to bootstrap, and the curl project doesn't maintain an ftp server.

Providing a script that prefetch (like what @natefoo started) would be a good compromise then:
* We get ride of `wget` as you suggested, 
* then we a standalone prefetch script that need to be called before running docker
* and we simplify the system expecting that the prefetch script is called beforehand

njsmith:
No. The problem is that the openssl in centos5 doesn't support SNI, but more and more servers are requiring that. If you look at the history of pull requests, we've been scrambling for months to switch between different mirrors for openssl and curl, trying to find ones that still work with centos5. This is the only reason we're using `ftp` in the first place.

So from my point of view, the whole point of switching to prefetching is that we no longer have play this game – we can use regular download sources like regular people, and stop having the image be broken constantly like it is now. However, centos5's `wget` can't connect to regular download sources, no matter which build service you're using, so I don't think that fallback is going to work. Sorry.

njsmith:
(edited my last comment to make it clear what I was responding to)

jcfr:
@natefoo Do you think you could you improve your last patch removing the support for wget and making the use of the prefetch a requirement ?  Let me know if I can help.

natefoo:
Well, it can connect to the official OpenSSL FTP source, but yeah, for curl it's using the Debian archive of the upstream source, so that one's not regular. I'm not disagreeing that it's a goose chase, and it will indeed continue to break going foward. I guess I was just hoping to retain the functionality for as long as it could be made to work. But, if you feel strongly that it should go, I'll rip out the remaining wget fetching and just switch to prefetches being required for OpenSSL and curl. Sound good?

jcfr:
Nice.

njsmith:
Looks good. Thanks a lot for taking care of this; it's not fun but it helps a lot.

jcfr:
> t's not fun but it helps a lot.

Different type of fun I would say.

Thanks for spending time reviewing and integrating :+1:

PR #186, opened 2018-04-20
remove cpython 3.3 support

trishankatdatadog:
update supported cpythons to latest versions
add convenience build script

Signed-off-by: Trishank K Kuppusamy <trishank.kuppusamy@datadoghq.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,8 +23,7 @@ matrix:
     - env: PLATFORM="x86_64"
 
 script:
-  - bash ./docker/build_scripts/prefetch.sh openssl curl
-  - docker build --rm -t quay.io/pypa/manylinux1_$PLATFORM:$TRAVIS_COMMIT -f docker/Dockerfile-$PLATFORM docker/
+  - PLATFORM=$(uname -m) TRAVIS_COMMIT=$TRAVIS_COMMIT ./build.sh
 
 
 deploy:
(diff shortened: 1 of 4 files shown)


njsmith:
I guess I'll leave this open for a few days in case anyone wants to make some vociferous argument for preserving 3.3 support...

I guess in a perfect world the other changes (adding `build.sh`, bumping the other python versions to the latest point release) would be separate PRs, but I'm not going to be fussy about it. Having added `build.sh` though, could we also make `travis.yml` use it, to reduce duplication?

njsmith:
Ping, can you update `.travis.yml` to use `build.sh`?

trishankatdatadog:
Ops, my bad, will get to this tomorrow!

trishankatdatadog:
Hmm, why doesn't it run the tests?

njsmith:
That is weird. Let me try doing the  closing/opening dance...

trishankatdatadog:
maybe because it's based on another PR? i'm not sure...

trishankatdatadog:
it's not the cleanest approach, but we could try merging with that PR, and see what happens...

njsmith:
I don't think that should make a difference... Maybe we should try opening a PR with some trivial change and see if that runs there tests, to check if it's just this PR or all PRs that are broken?

trishankatdatadog:
i can try sending another PR tomorrow?

njsmith:
Replaced by #188

PR #188, opened 2018-05-01
Update wheel from 0.30.0 to 0.31.0

trishankatdatadog:
- Remove CPython 3.3 support due to conflicting requirements, and
relative lack of support
- Update supported CPython versions
- Add convenience build script


njsmith:
Huh, so this version worked? Mysterious, but... I guess I'll take it :-). Thanks!

PR #182, opened 2018-04-16
Manylinux2010

dolang:
Now that there's some activity again here, I figure it's easier to expose the changes I made as a pull request so you can have a look at the diffs.

This is from my `manylinux2010` branch which is based on `markrwilliams:manylinx2` merged into `pypa:master`.

I tried to keep this branch close to upstream and just merged the most recent commit to master (4a20e4b) with minor adjustments.

I don't expect this to be merged by itself, but hope it saves you guys some time.  Notable:
- Removed `gpg` from `PYTHON_COMPILE_DEPS` because it's already present on CentOS 6 and its erasure later in the script breaks the build.
- Removed `ncurses-devel` as required by the PEP.
- Added `perl-devel`, seems to be required for the build now.
- Added a workaround for `libcurl.so` incompatibility (see commit notes & comments).
- Removed the `get-pip.py` workaround. That has been fixed (see [here](https://github.com/pypa/get-pip/issues/19)) and my build script works again.
- Moved `libffi-devel` from `PYTHON_COMPILE_DEPS` to the *"Development tools and libraries"* section.
- Changed the `x86_64` Docker tag from `:6.9` to `:6`. This mirrors a `manylinux1` change intended to reduce the overall image size (see pypa#206 and also [:6 Docker branch](https://github.com/CentOS/sig-cloud-instance-images/tree/CentOS-6/docker) / [:6.9 Docker branch](https://github.com/CentOS/sig-cloud-instance-images/tree/CentOS-6.9/docker)).
- Moved from CentOS 6.9 to 6.10
- **Careful:** I also changed occurrences of `manylinux1` to either `manylinux` or `manylinux2010`. This might be undesired.

~Side note: not a fan of the `/dev/null` out/error stream redirections in the scripts. They just make it harder to debug when something fails.  And there definitely are moving parts, as the `get-pip.py` problem has shown.~

PR #196, opened 2018-05-19
Add Python 3.7.0b4

lelit:
This is just to trigger a build and see how it goes.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -99,16 +99,18 @@ $ PLATFORM=$(uname -m) TRAVIS_COMMIT=latest ./build.sh
 
 Example
 -------
+
 An example project which builds 32- and 64-bit wheels for each Python interpreter
 version can be found here: https://github.com/pypa/python-manylinux-demo.
 
 This demonstrates how to use these docker images in conjunction with auditwheel
 to build manylinux-compatible wheels using the free `travis ci <https://travis-ci.org/>`_
-continuous integration service. 
+continuous integration service.
 
-(NB: for the 32-bit images running on a 64-bit host machine, it's necessary to run 
+(NB: for the 32-bit images running on a 64-bit host machine, it's necessary to run
 everything under the command line program `linux32`, which changes reported architecture
-in new program environment. See `this example invocation <https://github.com/pypa/python-manylinux-demo/blob/master/.travis.yml#L14>`_)
+in new program environment. See `this example invocation
+<https://github.com/pypa/python-manylinux-demo/blob/master/.travis.yml#L14>`_)
 
 The PEP itself
 --------------
--- a/README.rst
+++ b/README.rst
@@ -58,7 +58,7 @@ directory for source code.
 
 The images currently contain:
 
-- CPython 2.7, 3.4, 3.5 and 3.6, installed in
+- CPython 2.7, 3.4, 3.5, 3.6 and 3.7, installed in
   ``/opt/python/<python tag>-<abi tag>``. The directories are named
   after the PEP 425 tags for each environment --
   e.g. ``/opt/python/cp27-cp27mu`` contains a wide-unicode CPython 2.7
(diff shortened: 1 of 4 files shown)


lelit:
Uhm, this confuses me:

1. why does the build completes on TravisCI, and does not on my dev machine?
2. why it says "successful check", despite [this error](https://travis-ci.org/pypa/manylinux/jobs/380992812#L2152)?

njsmith:
Oh ick, yeah, that build totally failed, but Travis is reporting it as successful, which is worrisome. The problem is that `build.sh` doesn't check for failures. We need to add a `set -ex` or something to that top of that. (CC @trishankatdatadog)

Regarding the failure itself, it looks like we're failing to build the `ctypes` module for some reason, though I don't know why off the top of my head. To debug this, you might want to try commenting out the `>/dev/null` bits in the `do_cpython_build` function in `docker/build_scripts/build_utils.sh` – that should at least let you see any messages printed during the build. (We have to keep this disabled in the real version, b/c TravisCI blows up if you print too much output during a build.)

lelit:
I'm trying to understand why I'm not able to run the build on my local machine:
```
$ uname -a
Linux nautilus 4.16.0-1-amd64 #1 SMP Debian 4.16.5-1 (2018-04-29) x86_64 GNU/Linux
```

The exit status `139` seems related to a SEGV error, and indeed I'm not able to execute even a simple bash inside a centos:5.11 container:
```
$ docker run --rm -ti centos:5.11 /bin/bash
$
```

while I would expect something different, like:
```
$ docker run --rm -ti centos /bin/bash
[root@a6b8f02921be /]#
```

Will keep trying...

lelit:
No way to get it working on my Debian sid :sob: but at least on macOS it seems working, going to dig there.

lelit:
AFAICT, it fails building the ctypes extension because the related ffi libs/headers are missing...

lelit:
Maybe I should retry to build this on top of #182?

njsmith:
Oh, if you're getting an immediate segfault locally, that's a known issue: Debian is using a kernel with vsyscall=none. #158 is one solution, or possibly simpler is to edit your kernel command line to add `vsyscall=emulate`. (At least I think that's what it is; googling `vsyscall docker` should find more details.)

lelit:
Thanks for the enlightenment!

njsmith:
Okay, yeah, so the issue is that [starting in 3.7](https://docs.python.org/3.7/whatsnew/3.7.html#build-changes), CPython requires that a system copy of libffi be installed:

> A full copy of libffi is no longer bundled for use when building the _ctypes module on non-OSX UNIX platforms. An installed copy of libffi is now required when building _ctypes on such platforms. (Contributed by Zachary Ware in [bpo-27979](https://bugs.python.org/issue27979).)

So the first thing to try is adding `libffi-devel` to `PYTHON_COMPILE_DEPS` and see if that works any better.

lelit:
Yay, it seems working now!

lelit:
Should I upgrade this to b5, or for some reason is this a dead path?

mehrdada:
I am wondering what the ETA on merging this change is? We want to use the docker image to build Python3.7 gRPC packages.

hugovk:
It would be great to have this merged (with the [latest 3.7.0 candidate 1](https://www.python.org/dev/peps/pep-0537/#release-schedule), and updated later with 3.7.0 final) so downstream users can begin prep for the Python 3.7 release. 

For example, [Pillow](https://github.com/python-pillow/pillow-wheels/pull/86) is waiting for [multibuild](https://github.com/matthew-brett/multibuild) to add 3.7, which is waiting for manylinux to add it.

Thank you!

lelit:
I switched the setting to import 3.7.0rc1.

njsmith:
Sorry for being slow, I've just been juggling a lot of things lately. This looks great.

lelit:
Thank you Nathaniel!

PR #200, opened 2018-05-20
halt on error during build

trishankatdatadog:
addresses https://github.com/pypa/manylinux/pull/196#issuecomment-390386869

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+# Stop at any error, show all commands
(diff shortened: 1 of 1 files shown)


trishankatdatadog:
Thanks, @njsmith, very kind of you! Sure, as long as you're around, I'd love to :) I'm not an expert on everything (especially with the manylinux2010 stuff, haven't caught up with everything), but I should be able to help around.

Issue #202, opened 2018-05-23
wget https broken in manylinux images


dvarrazzo:
See [this build failure](https://travis-ci.org/psycopg/psycopg2-wheels/jobs/382886588):

```
+ wget https://github.com/openssl/openssl/archive/OpenSSL_1_0_2o.tar.gz
--2018-05-23 17:13:33--  https://github.com/openssl/openssl/archive/OpenSSL_1_0_2o.tar.gz
Resolving github.com... 192.30.253.113, 192.30.253.112
Connecting to github.com|192.30.253.113|:443... connected.
OpenSSL: error:1407742E:SSL routines:SSL23_GET_SERVER_HELLO:tlsv1 alert protocol version
Unable to establish SSL connection.
```

mayeut:
I guess `wget` is the system one, using the system `OpenSSL` which does not support a recent enough TLS protocol version. This is a common problem these days...
If you can switch to `curl`, it should fix your issue. It's been rebuilt from sources in the `manylinux` image using `OpenSSL 1.0.2o`.

dvarrazzo:
No wait: don't remove wget. curl gets stuck more than 50% of time on ftp. Sigh.

https://travis-ci.org/psycopg/psycopg2-wheels/builds/382907565

trishankatdatadog:
Does anyone else have reservations about this? The problem is that we don't know what this will break downstream until we break it. I think it's harmless to leave `wget` in unless it becomes a security problem.

mayeut:
This issue has been opened for almost a year, closing it. Please reopen if needed.

`wget` is not installed anymore on manylinux images. Please use `curl`.

PR #204, opened 2018-05-23
Remove wget utility

mayeut:
The system provided wget utility shall not be used, it's still using the system provided OpenSSL which is too old for most secure downloads (no TLS v1.1/v1.2 support).

### DIFF
--- a/docker/build_scripts/build.sh
+++ b/docker/build_scripts/build.sh
@@ -54,6 +54,9 @@ mv devtools-2.repo /etc/yum.repos.d/devtools-2.repo
 rpm -Uvh --replacepkgs epel-release-5*.rpm
 rm -f epel-release-5*.rpm
 
+# from now on, we shall only use curl to retrieve files
+yum -y erase wget
+
 # Development tools and libraries
 yum -y install \
     automake \


mayeut:
c.f. #202 for my motivation. I'd rather have no `wget` utility (with the obvious `curl` alternative working).
Another option would be to rebuild from sources, as is done for `curl`, in order to still provide the `wget` utility.

dvarrazzo:
fyi, with curl I'm experiencing an high download failure rate of an ftp resource. I don't know if it's their server's fault, or curl's, or if there is some timeout/retry switch on curl. The above change (getting the resource via https instead) seems working so far in my case.

mayeut:
@dvarrazzo, you're probably running into travis-ci/travis-ci#9391 when using FTP.

ehashman:
@dvarrazzo is your comment still blocking or have you been able to figure out a working solution with curl?

trishankatdatadog:
Does anyone else have reservations about this? The problem is that we don't know what this will break downstream until we break it. I think it's harmless to leave wget in unless it becomes a security problem.

mayeut:
Another option as mentioned in earlier comment is to rebuild wget from sources. I can probably update the PR to do just that.

trishankatdatadog:
@mayeut I'm not sure that's worth the effort, too. On second thought, if we break wget somewhere downstream, it's not the hardest thing to switch to curl. I think we can safely merge this.

PR #206, opened 2018-05-26
Reduce image size

mayeut:
1st commit fixes python test removal: around 100MB saved
2nd commit makes x86_64 image based on `centos:5` rather than `centos:5.11`: around 150MB saved

### DIFF
--- a/docker/Dockerfile-x86_64
+++ b/docker/Dockerfile-x86_64
@@ -1,4 +1,4 @@
-FROM centos:5.11
+FROM centos:5
 MAINTAINER The ManyLinux project
 
 ENV LC_ALL en_US.UTF-8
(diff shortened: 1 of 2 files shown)


trishankatdatadog:
@mayeut Could you please explain the difference between the two tags?

mayeut:
@trishankatdatadog,

centos:5.11 Stock image, identical to what you'd get from a fresh install using the 5.11 ISO
centos:5 base image has been built one year after the centos:5.11 image and thus already includes some package upgrades.
https://github.com/CentOS/sig-cloud-instance-images/tree/CentOS-5/docker
https://github.com/CentOS/sig-cloud-instance-images/tree/CentOS-5.11/docker

trishankatdatadog:
Got it, thanks @mayeut! As long as we always update the Centos within whenever we build, I think this is fine.

mayeut:
Rebased on up-to-date master which now includes python 3.7
```
quay.io/pypa/manylinux1_x86_64   pr       1aee74baba30    18 hours ago    881MB
quay.io/pypa/manylinux1_x86_64   latest   017377d5bcc4    37 hours ago    1.16GB
```

@trishankatdatadog, CentOS is still being updated on build. You can also find the list of installed packages along with their version in the build log. It's the exact same one with or without the PR.

trishankatdatadog:
LGTM

PR #207, opened 2018-05-28
Remove get-pip.py from build scripts now that curl support TLS v1.2

mayeut:
`curl` is rebuilt with TLS v1.2 support. `get-pip.py` doesn't need to be stored in build scripts anymore.


ehashman:
It looks like this script is still used in `docker/build_scripts/build_utils.sh`, could you look at updating that to work with the removal?

mayeut:
The bundled script is only used at line 96, which has changed as can be seen in the diff.
Other references of `get-pip.py` now only use the script retrieved with `curl` (still at line 96) unless I'm missing something.

trishankatdatadog:
This looks fine by me. Thanks @mayeut!

PR #210, opened 2018-07-07
[WIP] Reduce image size

mayeut:
1st & 2nd commits provide a small improvement and are quite straightforward.
The 3rd commit allows around 60MB to be saved. It uses a shared library of OpenSSL rather than the static version. OpenSSL & curl shared libraries are isolated from system ones using patchelf to change soname and dependencies (as does auditwheel for manylinux wheels).
Overall, 74MB saved.


mayeut:
@dolang, you might be interested in the 3rd commit for #182. curl & openssl shared libraries should not interfere with system ones. This allows to remove the hack around `yum` in #182 while reducing image size.

mayeut:
While working on a similar PR for manylinux2010, I found some minor quirks that need resolving.
Work In Progress

mayeut:
Let's no waste energy on manylinux1, I close this for now.

PR #229, opened 2018-10-01
Revert update to wheel 0.32

matthew-brett:
Sorry about this - but Wheel 0.32 breaks auditwheel, and as auditwheel doesn't
have tests for the relevant code, I suspect it will be a while before it works
with 0.32.0 - see https://github.com/pypa/auditwheel/issues/102

So, can I humbly ask that we revert to Wheel 0.3.1, in order to unbreak
auditwheel, in the meantime?


njsmith:
It's unfortunate that our CI didn't notice any issue with #228... is there anything we could do to catch this kind of issue in the future?

matthew-brett:
I guess we could try building a wheel that needs repairing in the tests.

trishankatdatadog:
Sorry, guys, my bad

PR #233, opened 2018-10-08
Update build_utils.sh to leave python headers in place

bjmuld:
I'm building python extensions which rely on boost-python.... boost python has to be built against python headers.... using --enable-shared at python build-time leaves the necessary python headers in place for each python release....

### DIFF
--- a/docker/build_scripts/build_utils.sh
+++ b/docker/build_scripts/build_utils.sh
@@ -44,7 +44,7 @@ function do_cpython_build {
     fi
     local prefix="/opt/_internal/cpython-${py_ver}${dir_suffix}"
     mkdir -p ${prefix}/lib
-    ./configure --prefix=${prefix} --disable-shared $unicode_flags > /dev/null
+    ./configure --prefix=${prefix} --enable-shared $unicode_flags > /dev/null
     make -j2 > /dev/null
     make install > /dev/null
     popd


bjmuld:
I should add that I don't think there are centos-5 packages for python3-devel... there is python-devel, but why not just leave the headers in place within each python dist?

rdb:
The Python headers *are* in the image already, so it's not clear what this PR is intended to accomplish.  It seems to rather build Python as a shared library, which seems quite undesirable and has nothing to do with the presence of the headers.

bjmuld:
is that so?  I suppose I should have done more checking before the request... Boost's boostrap script is finding headers for python27 but can't find python36 or python 37....  I guess I'll  have to give it a hand.

bjmuld:
can I ask what the "m" suffix indicates?  Boost looks for --and finds-- those include dirs which don't have the suffix (is there a case for consistency?):

/opt/python/cp27-cp27mu/include/python2.7/pyconfig.h
/opt/python/cp37-cp37m/include/python3.7m/pyconfig.h
/opt/python/cp35-cp35m/include/python3.5m/pyconfig.h
/opt/python/cp27-cp27m/include/python2.7/pyconfig.h
/opt/python/cp34-cp34m/include/python3.4m/pyconfig.h
/opt/python/cp36-cp36m/include/python3.6m/pyconfig.h

auvipy:
closing for now

PR #252, opened 2018-12-11
[WIP] manylinux2010 with devtoolset 7

henriquegemignani:
I've rebased #152 on top of a recent master because I needed to create a manylinux2010 wheel for Python 3.7.
I also needed a recent toolchain so I installed the devtoolset 7 from The Software Collections.

This PR is not ready at all for merge, especially compared to #182, but I feel it's good enough for comparison and discussion.

* I took another approach to the curl issue: install to `/opt` and build python/git with it only.
* i686 wasn't working with devtoolset 7


Mizux:
My two cents,
You should bump auditwheel to `2.0.0rc1` otherwise auditwheel doesn't understand `manylinux2010` policy...

BTW thanks for your [docker image](https://hub.docker.com/r/henriquegemignani/manylinux/tags) :+1: (can you build it, so it's work without vsyscall ? ;))

mayeut:
Superseded by #279 that's been merged-in.

PR #279, opened 2019-03-19
[WIP] Manylinux2010

mayeut:
A new Manylinux 2010 PR based on #182 with master merged c.f. https://github.com/pypa/manylinux/issues/179#issuecomment-473689088

Remaining items:

- [ ] Thorough  review (already began in #182)
- [ ] Merge #210 in master (or at least the commit allowing to get rid of yum workaround) then rebase. This requires some review. This leads to yum workaround avoided, proper isolation of curl & openssl and a reduced image (faster docker pull). This could be done in a second step but this is the occasion to get this merged in.
- [x] drop 32bit image, update to devtoolset-8


docker/build_scripts/build.sh
```diff
@@ -10,20 +10,10 @@ MY_DIR=$(dirname "${BASH_SOURCE[0]}")
...
-# Centos 5 is EOL and is no longer available from the usual mirrors, so switch
-# to http://vault.centos.org
-# From: https://github.com/rust-lang/rust/pull/41045
-# The location for version 5 was also removed, so now only the specific release
-# (5.11) can be referenced.
-sed -i 's/enabled=1/enabled=0/' /etc/yum/pluginconf.d/fastestmirror.conf
-sed -i 's/mirrorlist/#mirrorlist/' /etc/yum.repos.d/*.repo
-sed -i 's/#\(baseurl.*\)mirror.centos.org\/centos\/$releasever/\1vault.centos.org\/5.11/' /etc/yum.repos.d/*.repo
+# Libraries that are allowed as part of the manylinux2010 profile
+MANYLINUX2010_DEPS="glibc-devel libstdc++-devel glib2-devel libX11-devel libXext-devel libXrender-devel mesa-libGL-devel libICE-devel libSM-devel"
```
illume: Here is the list of allowed shared executables in the spec: https://www.python.org/dev/peps/pep-0571/#id49

docker/build_scripts/build.sh
```diff
@@ -41,39 +31,28 @@ echo "multilib_policy=best" >> /etc/yum.conf
...
-# Dev toolset (for LLVM and other projects requiring C++11 support)
-wget -q http://people.centos.org/tru/devtools-2/devtools-2.repo
-check_sha256sum devtools-2.repo $DEVTOOLS_HASH
-mv devtools-2.repo /etc/yum.repos.d/devtools-2.repo
-rpm -Uvh --replacepkgs epel-release-5*.rpm
-rm -f epel-release-5*.rpm
-
-# from now on, we shall only use curl to retrieve files
-yum -y erase wget
+# Software collection and EPEL support
```
illume: For anyone interested... What's an EPEL for? https://fedoraproject.org/wiki/EPEL?rd=EPEL/en

docker/build_scripts/manylinux-check.py
```diff
@@ -1,6 +1,6 @@
 # Logic copied from PEP 513
```
illume: Should this file be named `docker/build_scripts/manylinux-check2010.py` ?

docker/build_scripts/build.sh
```diff
@@ -147,18 +126,40 @@ rm -rf patchelf.tar.gz patchelf-$PATCHELF_VERSION
 
 ln -s $PY36_BIN/auditwheel /usr/local/bin/auditwheel
 
+# HACK: The newly compiled and installed curl messes with the system's
+# py2.6 installation, on which yum depends.  Work around it by
+# rewiring libcurl.so specifically for yum.  /usr/local/bin/ has higher
+# priority on the PATH than /usr/bin/
+cat <<'EOF' > /usr/local/bin/yum && chmod +x /usr/local/bin/yum
+#!/bin/bash
+if [ "x$(arch)" != xi686 ]; then
```
illume: This test should probably be more specific? So that it fails hard with other arch's like arm32.

docker/glibc/build_scripts/CentOS-source.repo
```diff
@@ -0,0 +1,7 @@
+[base-source]
+name=CentOS-6.10 - Base SRPMS
+baseurl=http://vault.centos.org/6.10/os/Source/
+gpgcheck=1
+gpgkey=http://mirror.centos.org/centos/RPM-GPG-KEY-CentOS-6
```
illume: URL for gpgkey is wrong (and http).
See here: https://www.centos.org/keys/#centos-6-signing-key

links to here: https://www.centos.org/keys/RPM-GPG-KEY-CentOS-6

docker/glibc/build_scripts/glibc.spec.patch
```diff
@@ -0,0 +1,29 @@
...
+ Buildroot: %{_tmppath}/%{name}-%{version}-%{release}-root-%(%{__id_u} -n)
+ Obsoletes: glibc-profile < 2.4
+@@ -731,6 +732,7 @@
+ %patch247 -p1
+ %patch248 -p1
+ %patch249 -p1
++%patch250 -E -p3
+ 
+ # A lot of programs still misuse memcpy when they have to use
+ # memmove. The memcpy implementation below is not tolerant at
```
illume: This sentence doesn't seem to finish?

docker/build_scripts/build.sh
```diff
@@ -10,20 +10,10 @@ MY_DIR=$(dirname "${BASH_SOURCE[0]}")
 
 # Dependencies for compiling Python that we want to remove from
 # the final image after compiling Python
-# GPG installed to verify signatures on Python source tarballs.
-PYTHON_COMPILE_DEPS="zlib-devel bzip2-devel ncurses-devel sqlite-devel readline-devel tk-devel gdbm-devel db4-devel libpcap-devel xz-devel gpg libffi-devel"
+PYTHON_COMPILE_DEPS="zlib-devel bzip2-devel ncurses-devel sqlite-devel readline-devel tk-devel gdbm-devel db4-devel libpcap-devel xz-devel"
```
illume: expat, and openssl are not specified here?

docker/glibc/build_scripts/remove-vsyscall.patch
```diff
@@ -0,0 +1,401 @@
+diff --git a/BUILD/glibc-2.12-2-gc4ccff1/nptl/sysdeps/unix/sysv/linux/x86_64/lowlevellock.S b/BUILD/glibc-2.12-2-gc4ccff1/nptl/sysdeps/unix/sysv/linux/x86_64/lowlevellock.S
```
illume: Who wrote this patch?

I found this commit... is it from here?  https://github.com/markrwilliams/manylinux/commit/e9493d55471d153089df3aafca8cfbcb50fa8093#diff-3eda4130bdba562657f3ec7c1b3f5720

I'm not sure who wrote it, if they need credit, or if it needs review.

docker/glibc/README.rst
```diff
@@ -0,0 +1,79 @@
+centos-6-no-vsyscall
+====================
+
+*Summary*: Because of
+https://mail.python.org/pipermail/wheel-builders/2016-December/000239.html,
+this a CentOS 6.10 Docker image that rebuilds ``glibc`` without
+*vsyscall* is necessary to reliably run ``manylinux2`` on 64-bit
```
takluyver: manylinux2 -> manylinux2010 throughout this document

docker/build_scripts/build.sh
JonasVautherin: Looking at the list provided, it seems that this is accurate, isn't it? `libgcc` does not appear here (but neither did it on manylinux1, so I guess it comes as a dependency of the others) and `libSM` is not listed in the packages, but appears in the libs as `libSM.so.6`.

docker/glibc/README.rst
mayeut: Done, thanks @JonasVautherin

docker/build_scripts/build.sh
illume: I don't see `libgcc: | libgcc_s.so.1` but I guess this provided below by "devtoolset-7-gcc" which is also installed.

mayeut: I added a comment that shall answer all questions here.
The list is accurate.

docker/build_scripts/build.sh
mayeut: Added a comment on what they used for

docker/build_scripts/manylinux-check.py
mayeut: Not sure it should be, that just another thing to change on the next update. I'd rather keep it that as it is.

docker/build_scripts/build.sh
mayeut: That's one of the reason I'd like to get #210 merged in, remove this hack once & for all. I'd rather do some rework on #210 which has some added value rather than rework this hack.

docker/build_scripts/build.sh
mayeut: openssl is a rebuilt from sources, not using centos package for that.
I moved expat-devel  to PYTHON_COMPILE_DEPS (was installed but not mentioned here)

docker/glibc/build_scripts/remove-vsyscall.patch
mayeut: I guess @markrwilliams wrote it ? in #152

docker/glibc/build_scripts/glibc.spec.patch
mayeut: It's a patch, that's expected.

docker/glibc/build_scripts/CentOS-source.repo
mayeut: Now using local gpgkey

docker/glibc/build_scripts/remove-vsyscall.patch
markrwilliams: I did, largely backporting vsyscall-less code from newer glibcs.

.travis.yml
```diff
@@ -19,11 +19,10 @@ env:
 
 matrix:
   include:
-    - env: PLATFORM="i686"
     - env: PLATFORM="x86_64"
 
 script:
-  - PLATFORM=$PLATFORM TRAVIS_COMMIT=$TRAVIS_COMMIT ./build.sh
```
takluyver: Is using `travis_wait` unavoidable? It appears to hide all output until the command completes - and if the build hits the time limit (e.g. [this one](https://travis-ci.org/pypa/manylinux/builds/514342665)), you never see any output at all, so you can't see how far it got or if something was hanging.

mayeut: I'm trying something found there: https://github.com/travis-ci/travis-ci/issues/4190

mayeut: Ok, we have logs but build still times out.
The build is passing on my end... https://travis-ci.com/mayeut/manylinux/builds/107284605

mayeut: Now a 2 build stage with cache, logs are here and build build time can be over 50 minutes.

takluyver:
I had to restart a timed-out job before, so I guess it was right on the cusp of the 50 minute timeout. If splitting it into two stages gives us 50 minutes for each stage, that should make it much more reliable.

takluyver:
@mayeut what level of confidence would you have in images built by this branch as it stands? There don't seem to be a lot of people stepping up to do review. Would you be happy if it were merged, so the images are posted to quay.io and people start using them, and we try to fix problems as they come up? Or do you think we should keep it on hold for further review?

JonasVautherin:
I cannot review this, but I am building stuff with manylinux1 and I could definitely try with manylinux2010 and report issues.

PR #292, opened 2019-04-17
Add an AUDITWHEEL_PLAT environment variable

mayeut:
The idea is that this environment variable could be used as a default value to the --plat argument of auditwheel repair command


mayeut:
c.f. matthew-brett/multibuild#238

PR #303, opened 2019-04-28
Don't rebuild curl

mayeut:
yum and other tools from the system rely on specific features provided by the system libcurl.so.4
A workaround exists only for yum but other tools are also failing.

curl was rebuilt on CentOS 5 in order to get TLS 1.2 support.
There's no need to rebuilt it on CentOS 6 to get TLS 1.2 support which is native.


njsmith:
I wonder if we should stop building our own curl/openssl now altogether. We only started doing that because the versions on centos5 got so old that they gave errors when connecting to modern web sites.

mayeut:
If we don't need to support TLS 1.3 in the near future and using CentOS 6 provided OpenSSL after CentOS 6 EOL (November 30th, 2020) doesn't pose a critical security threat then sure, I think you're right.
TLS 1.3 is still pretty young so I think we might be good on that specific need.
What about CentOS 6 going EOL ? (Hopefully we'll have something new by then ?) If you do think we won't need an updated OpenSSL then I'll update the PR not to build OpenSSL and curl altogether.

trishankatdatadog:
I concur with @njsmith

mayeut:
good, I'll rework that PR to get rid of openssl/curl build then.

njsmith:
The reason we needed a new libssl was the PCI council forcing everyone onto TLS 1.2 last year. TLS 1.2 was standardized in 2008. So if that's the pattern, we've got about a decade before TLS 1.3 becomes mandatory :-)

mayeut:
Seems like python 3.7 is not compatible with the OpenSSL 1.0.1e provided by the system:
c.f. https://docs.python.org/3/whatsnew/3.7.html#ssl
```
/Python-3.7.3 /
+ pushd Python-3.7.3
+ '[' none = none ']'
+ unicode_flags=
+ dir_suffix=
+ local prefix=/opt/_internal/cpython-3.7.3
+ mkdir -p /opt/_internal/cpython-3.7.3/lib
+ ./configure --prefix=/opt/_internal/cpython-3.7.3 --disable-shared
+ make -j2
+ make install
/
+ popd
+ rm -rf Python-3.7.3
+ '[' -e /opt/_internal/cpython-3.7.3/bin/python3 ']'
+ ln -s python3 /opt/_internal/cpython-3.7.3/bin/python
+ /opt/_internal/cpython-3.7.3/bin/python get-pip.py
WARNING: pip is configured with locations that require TLS/SSL, however the ssl module in Python is not available.
Collecting pip
  WARNING: Retrying (Retry(total=4, connect=None, read=None, redirect=None, status=None)) after connection broken by 'SSLError("Can't connect to HTTPS URL because the SSL module is not available.")': /simple/pip/
  WARNING: Retrying (Retry(total=3, connect=None, read=None, redirect=None, status=None)) after connection broken by 'SSLError("Can't connect to HTTPS URL because the SSL module is not available.")': /simple/pip/
  WARNING: Retrying (Retry(total=2, connect=None, read=None, redirect=None, status=None)) after connection broken by 'SSLError("Can't connect to HTTPS URL because the SSL module is not available.")': /simple/pip/
  WARNING: Retrying (Retry(total=1, connect=None, read=None, redirect=None, status=None)) after connection broken by 'SSLError("Can't connect to HTTPS URL because the SSL module is not available.")': /simple/pip/
  WARNING: Retrying (Retry(total=0, connect=None, read=None, redirect=None, status=None)) after connection broken by 'SSLError("Can't connect to HTTPS URL because the SSL module is not available.")': /simple/pip/
  Could not fetch URL https://pypi.org/simple/pip/: There was a problem confirming the ssl certificate: HTTPSConnectionPool(host='pypi.org', port=443): Max retries exceeded with url: /simple/pip/ (Caused by SSLError("Can't connect to HTTPS URL because the SSL module is not available.")) - skipping
  ERROR: Could not find a version that satisfies the requirement pip (from versions: none)
ERROR: No matching distribution found for pip
WARNING: pip is configured with locations that require TLS/SSL, however the ssl module in Python is not available.
Could not fetch URL https://pypi.org/simple/pip/: There was a problem confirming the ssl certificate: HTTPSConnectionPool(host='pypi.org', port=443): Max retries exceeded with url: /simple/pip/ (Caused by SSLError("Can't connect to HTTPS URL because the SSL module is not available.")) - skipping
The command '/bin/sh -c bash build_scripts/build.sh && rm -r build_scripts' returned a non-zero code: 1
```

njsmith:
Oh bleh, that's plausible.

mayeut:
I can still remove the prefetch, not rebuild curl to bypass the specific issue. OpenSSL still needs to be rebuilt for cpython but that's still one dependency we don't have to manage anymore.

mayeut:
Prefetch removed, custom build of curl removed, git now builds using system libraries only, openssl still custom built for cpython only.

ofek:
Can this be merged now?

PR #314, opened 2019-06-13
Add Python-3.8.0b1

filipArena:
It would make sense to add the beta before the release lands, so that package maintainers have time to adjust their pipelines and build wheels for 3.8.

### DIFF
--- a/docker/build_scripts/build_env.sh
+++ b/docker/build_scripts/build_env.sh
@@ -1,7 +1,7 @@
 # source me
 
 PYTHON_DOWNLOAD_URL=https://www.python.org/ftp/python
-CPYTHON_VERSIONS="2.7.16 3.4.10 3.5.7 3.6.8 3.7.3"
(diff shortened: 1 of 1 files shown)


njsmith:
The problem is, building wheels for 3.8b1 isn't very useful if those wheels crash when you use them on 3.8-final :-)

@ambv When does the 3.8 ABI get frozen so people can start building wheels?

auvipy:
after beta 3

lkollar:
We could create an image with `beta` or `experimental` tag for folks who want to experiment with the new release until it stabilises. Otherwise I agree that we should wait until the ABI becomes stable to avoid ABI-incompatible wheels in the wild.

njsmith:
The ideal for experimentation would be if we could arrange for pre-release wheels to get some unique tag, like `cp38b1` or something. This would require some assistance from pip though to make these wheels installable...

filipArena:
It seems that b1 isn't even a part of https://www.python.org/static/files/pubkeys.txt so this won't work even if the ABI gets stable. Maybe create another docker image for betas and skip gpg verify?

PR #325, opened 2019-07-19
manylinux2010: Install libxcrypt

lkollar:
This will make `libcrypt.so.2` available in the image. The GLIBC variants of the header and libraries are deleted in order to force applications to pick up the `libxcrypt` version.

See https://github.com/pypa/manylinux/issues/305.


lkollar:
When I tested the `libxcrypt` build locally in the `manylinux2010` image it failed with this error:

```
lib/crypt-md5.c: In function ‘_crypt_crypt_md5crypt_rn’:
lib/crypt-md5.c:86:15: error: conversion to ‘long unsigned int’ from ‘int’ may change the sign of the result [-Werror=sign-conversion]
   salt_size = strspn (salt, b64t);
               ^~~~~~
cc1: all warnings being treated as errors
make[1]: *** [lib/libcrypt_la-crypt-md5.lo] Error 1
make: *** [all] Error 2
lib/crypt-md5.c: In function ‘_crypt_crypt_md5crypt_rn’:
lib/crypt-md5.c:86:15: error: conversion to ‘long unsigned int’ from ‘int’ may change the sign of the result [-Werror=sign-conversion]
   salt_size = strspn (salt, b64t);
               ^~~~~~
cc1: all warnings being treated as errors
make: *** [lib/libcrypt_la-crypt-md5.lo] Error 1
```
GCC: 
```
[root@40d2f86f100f build_scripts]# gcc --version
gcc (GCC) 8.2.1 20180905 (Red Hat 8.2.1-3)
```
It is supposed to compile with GCC 9 so not sure if it's this particular GCC version or something else.

zackw:
Upstream libxcrypt uses extremely paranoid warnings settings for development, but unfortunately that means it's sensitive to the exact compiler version in use.  For installation into manylinux I recommend configuring with `--disable-werror`.

lkollar:
Successfully tested this with `pyphash`:

```
[root@9f14cc2348da ~]# /opt/python/cp37-cp37m/bin/python -c "import pyphash"
[root@9f14cc2348da ~]# ldd /opt/python/cp37-cp37m/lib/python3.7/site-packages/pyphash.cpython-37m-x86_64-linux-gnu.so
	linux-vdso.so.1 =>  (0x00007ffec37ef000)
	libcrypt.so.2 => /usr/local/lib/libcrypt.so.2 (0x00007f4a33f14000)
	libpthread.so.0 => /lib64/libpthread.so.0 (0x00007f4a33cf7000)
	libc.so.6 => /lib64/libc.so.6 (0x00007f4a33963000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f4a3434f000)
[root@9f14cc2348da ~]#
```

takluyver:
Don't forget to remove the tarball from this PR too.

lkollar:
Looks like I forgot to push the change to remove the tarball. Updated the PR.

PR #333, opened 2019-08-08
include 3.8.0b3

### LINKED ISSUE #313: Python 3.8
Issue to track adding Python 3.8 to the images, which it looks like should happen at the end of July, with 3.8.0b3.

From the [release notes for 3.8b1](https://www.python.org/downloads/release/python-380b1/), 3.8 ABI is planned to be frozen in beta 3, which is [scheduled](https://www.python.org/dev/peps/pep-0569/) for 2019-07-29.

Or maybe the policy in this repo should be freezing the ABI with release candidates? That would give a much shorter window for developers to start publishing wheels with this tool (3 weeks vs 3 months).

minrk:
beta 3 is meant to freeze the ABI, so this could be the right time.

closes #314
closes #313

### DIFF
--- a/docker/build_scripts/build_env.sh
+++ b/docker/build_scripts/build_env.sh
@@ -1,7 +1,7 @@
 # source me
 
 PYTHON_DOWNLOAD_URL=https://www.python.org/ftp/python
-CPYTHON_VERSIONS="2.7.16 3.4.10 3.5.7 3.6.9 3.7.4"
+CPYTHON_VERSIONS="2.7.16 3.4.10 3.5.7 3.6.9 3.7.4 3.8.0b3"
 
 # openssl version to build, with expected sha256 hash of .tar.gz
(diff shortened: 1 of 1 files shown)


takluyver:
It's failing on checking the GPG signature - as someone noted on #314, it's apparently not signed with a key in https://www.python.org/static/files/pubkeys.txt . I don't know whether that's something we should work around for the beta, or an oversight when making the release.

mattip:
ping. It would be nice to be able to start testing 3.8. Is there an upstream issue for resolving this?

mattip:
Filed [bpo-37967](https://bugs.python.org/issue37967)

mattip:
[3.8b4](https://www.python.org/downloads/release/python-380b4/) is out, could we try again with that one?

mattip:
Closed by #344

PR #358, opened 2019-10-01
fix pep8 complains

hrw:
manylinux-check.py:3:1: E302 expected 2 blank lines, found 1
manylinux-check.py:20:1: E302 expected 2 blank lines, found 1
manylinux-check.py:47:1: E305 expected 2 blank lines after class or function definition, found 1
manylinux-check.py:47:1: E402 module level import not at top of file
python-tag-abi-tag.py:7:75: W292 no newline at end of file
ssl-check.py:6:1: E402 module level import not at top of file


hrw:
My editor does PEP8 checks on load/save so I decided to fix reported issues so I can edit them without being reminded.

PR #348, opened 2019-09-22
manylinux2014 images

mayeut:
1st attempt at a manylinux2014 image
main differences with the 2010 image:
- drop python 2.7 (EOL in january 2020) & 3.4 (already EOL)
- use system provided openssl for python build
- cmake is not installed anymore (one can use pip install cmake)
- add aarch64 support
- i686 uses default gcc 4.8.5 & misses yasm

c.f. #338 

Please do comment

PR #362, opened 2019-10-07
WIP - Manylinux2014_aarch64

mayeut:
This builds on top of #348 to try to get travis-ci to build aarch64 without polluting #348, expect a lot of fix commits for now. 
I can't do this on my fork which use `.com` travis-ci and not `.org`. Although the doc states that it should be working on `.com`, it's not...


mayeut:
It's passing, I will update #348 instead.

PR #370, opened 2019-10-14
update for cpython 3.8.0 final

mattip:
update the version and also try to document the supported format for the next release cycle


ax3l:
@mattip @mayeut should the `README` be updated as well to reflect Python 3.8 support?

PR #378, opened 2019-10-21
Remove external typing package for python 3.7

### LINKED ISSUE #377: Package 'typing' installed in python >=3.5
The `typing` package seems to be installed in the manylinux python environments, even for python newer than python 3.5. It is my understanding that this package is a backport for the internal typing module for python versions older than 3.5, [right](https://pypi.org/project/typing/)?

```
$ docker run --rm quay.io/pypa/manylinux1_x86_64 bash -c "ls -l /opt/python/cp37-cp37m/lib/python3.7/site-packages/typing.py"
-rw-r--r-- 1 root root 80432 Oct 16 20:51 /opt/python/cp37-cp37m/lib/python3.7/site-packages/typing.py
```

I'm sure this is normally not an issue, but it has started causing issues for my CI where I use cibuildwheel (see related issue: https://github.com/joerick/cibuildwheel/issues/176). Regardless of whether or not this works for people, shouldn't this package *not* be installed for newer python versions? I would test my CI against older images but only the latest image is available from what I can tell.

djhoese:
Closes #377 

As mentioned in the `typing` package's documentation, it is meant for older versions of python. The documentation also says is should have NO EFFECT in newer versions of python, this is only true in some cases. The builtin python typing module works fine when called with subprocess, the `typing` package does not. See #377 for more details.

Edit: Should have pointed out that the recommended fix for the error discussed in #377 is to not install the typing package.

### DIFF
--- a/docker/build_scripts/py37-requirements.txt
+++ b/docker/build_scripts/py37-requirements.txt
@@ -9,8 +9,3 @@ auditwheel==2.1.1 \
 # this package required for auditwheel
 pyelftools==0.25 \
     --hash=sha256:89c6da6f56280c37a5ff33468591ba9a124e17d71fe42de971818cbff46c1b24
-# this package required for auditwheel
-typing==3.7.4.1 \
-    --hash=sha256:91dfe6f3f706ee8cc32d38edbbf304e9b7583fb37108fef38229617f8b3eba23 \
-    --hash=sha256:c8cabb5ab8945cd2f54917be357d134db9cc1eb039e59d1606dc1e60cb1d9d36 \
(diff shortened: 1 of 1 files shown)


djhoese:
Thanks @lkollar! It looks like the `manylinux2010_x86_64` image was updated. Does the `manylinux1_x86_64` image get pushed manually? I'm using a tool called cibuildwheel which I think defaults to the manylinux1 image still (waiting on the release for it to use 2010). Would it be possible to update the manylinux1 image or has it been officially deprecated and won't be updated anymore?

lkollar:
Yeah, `manylinux1` needs an update too. Submitted https://github.com/pypa/manylinux/pull/379 for it.

PR #385, opened 2019-10-29
Update setuptools from 41.4.0 to 41.6.0 (manylinux1)

mayeut:
cherry-picked from master


trishankatdatadog:
It's getting a bit tiresome to maintain different versions of manylinux, no? I'm not sure how long we should keep doing this, but thanks for keeping up!

mayeut:
> It's getting a bit tiresome to maintain different versions of manylinux, no?

 Yes a bit... It's still heavily used compared to manylinux2010 so I guess we'll have to continue maintaining this branch for a bit longer. That being said, I think I'll stop spending my time to keep things up-to-date in this branch after python 2.7 EOL or not on a regular basis at least. If a small update PR is opened by someone, I'll review/merge it.

PR #388, opened 2019-11-11
Add CPython 2.7.17

mayeut:
Nothing prevents adding CPython 2.7 so adding it.


mayeut:
Forgot about the different flavors...

PR #390, opened 2019-11-13
Add manylinux2014_ppc64le image

mayeut:
Travis-ci now provides ppc64le support, let's propose a ppc64le image.

Building autoconf from sources failed but the system provided one is the latest release (+ patch from CentOS) so not building autoconf from sources anymore.


mayeut:
If CI passes, README still needs to be updated.

PR #391, opened 2019-11-13
add s390x support

mattip:
Travis recently added s390x support, which appears as an option in the [PEP 599 manylinux2014](https://www.python.org/dev/peps/pep-0599/#the-manylinux2014-policy) spec. Add a build for it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,8 @@ matrix:
      env: PLATFORM="aarch64"
    - arch: ppc64le
      env: PLATFORM="ppc64le"
+   - arch: s390x
+     env: PLATFORM="s390x"
 
 script:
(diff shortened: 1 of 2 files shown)


auvipy:
build is failing

mattip:
build is still failing

mattip:
I guess I was optimistic that it would Just Work. The build starts, but cannot build the docker file. The first line in the docker file `FROM s390x/centos:7` fails with
`pull access denied for s390x/centos, repository does not exist or may require 'docker login'`

mayeut:
@mattip, thanks for trying and expressing interest in having this image available.
I opened PR #392 which should pass CI.
Closing this one.

PR #392, opened 2019-11-14
Add s390x support

mayeut:
Travis-ci now provides s390x support, let's propose a s390x image.

### DIFF
--- a/docker/Dockerfile-s390x
+++ b/docker/Dockerfile-s390x
@@ -1,0 +1,19 @@
+FROM s390x/clefos:7
+LABEL maintainer="The ManyLinux project"
+
+ENV AUDITWHEEL_ARCH s390x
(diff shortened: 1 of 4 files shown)


docker/Dockerfile-s390x
```diff
@@ -0,0 +1,19 @@
+FROM s390x/clefos:7
```
mattip: This is what was missing to get gh-391 going. For future reference, how did you know what to specify here?

mayeut: > For future reference, how did you know what to specify here?

Well, I didn't know. I googled around, found this: http://containerz.blogspot.com/2017/07/an-overview-on-s390x-base-images.html which lead me there https://hub.docker.com/_/clefos to gather the fact that the image was part of docker official images (i.e. good enough to use as a base image without too much further research) and https://www.sinenomine.net/products/linux/clefos (official page, mostly to gather if devtoolset-8 was part of the package list or not).

Issue #338, opened 2019-08-16
Tracking issue for manylinux2014 rollout


brainwane:
Now that [PEP 599 has been accepted](https://www.python.org/dev/peps/pep-0599/) to adopt the `manylinux2014` standard, we need a tracking issue for implementation (I'm reusing structure and prose from #179).

The dates are guesses but would (as I figure) help us get all the infrastructure in place well before Nov 2020 when CentOS 6 (supported in `manylinux2010`) will go EOL, and in time for [the August 2020 date](https://discuss.python.org/t/the-next-manylinux-specification/1043/150) that @gunan asked for; please argue with me. ;-)

There are a number of further steps needed before folks can actually make use of that new baseline, and the order of operations matters (since building `manylinux2014` wheels isn't very useful if installers won't install them):

#### Prep (ideally done by 2 September 2019):

- [ ]  Gather personnel: tentatively strategize re: who's available to review & merge code, especially for repositories where maintainer time is scarce, re: who will make a build environment, and re: who will do testing or build testing infrastructure
- [ ] Curate initial list of test cases regarding build environment, target Linux distributions, specific Python packages to be built, presence/absence of specific libraries

#### Essential client support (ideally done by 15 September 2019):
- [x]  Update `pip` 19.x to install `manylinux2014` wheels
  - [x] Update `packaging.tags` to detect `manylinux2014` compatibility (https://github.com/pypa/packaging/pull/186)
  - [x] New release of `packaging` ([19.2](https://pypi.org/project/packaging/19.2/))
  - [x] Add `manylinux2014` support to `pip._internal.pep425tags` (https://github.com/pypa/pip/pull/7102/)
  - [x] (Optional) Replace `pip._internal.pep425tags` w/ `packaging.tags` (https://github.com/pypa/pip/issues/6908)
  - [x] New release of `pip` ([19.3](https://github.com/pypa/pip/issues/6993) due out in week 41, so, by 19 October)

#### Enable publication of `manylinux2014` wheel archives (ideally done by 30 October 2019):
- [x]  Update `Warehouse` (PyPI) to accept `manylinux2014` wheel uploads (https://github.com/pypa/warehouse/pull/6684)
- [x]  Update `auditwheel` to emit `manylinux2014` wheel archives (https://github.com/pypa/auditwheel/pull/192)
  - [x]  Update `auditwheel` to allow new architectures introduced with `manylinux2014`  (https://github.com/pypa/auditwheel/pull/202)
[...]

PR #399, opened 2019-11-27
List all currently available docker images

mattip:
xref [this comment](https://github.com/pypa/manylinux/issues/338#issuecomment-558500000) in gh-338 where it was suggested to catalogue all available images.


mattip:
fixed, thanks

Issue #427, opened 2020-01-13
Drop support for CPython 3.4 (manylinux1/manylinux2010)


mayeut:
CPython 3.4 has reached end-of-life in march 2019

rdb:
This need not be a reason to keep support, but I'm just pointing out for the record that Ubuntu Trusty ships with Python 3.4 and won't be EOL until April 2022.

mayeut:
More details on Trusty EOL:
"standard" support for Trusty ended in April 2019.
There is in fact an extended paid support until April 2022 and thus EOL is mentioned as April 2022.
https://wiki.ubuntu.com/Releases

Python 3.4 keeps us from updating some python packages (or we'd need a specific requirement file) as well as upgrading OpenSSL (1.0.2 series is EOL) leading to potential security issues.

lkollar:
I second this. We have several version upgrade PRs failing because of Python 3.4. We should just remove it.

mayeut:
Reopen => manylinux1

PR #438, opened 2020-01-13
Drop python 3.4 support

mayeut:
CPython 3.4 has reached end-of-life in march 2019
Relates to #427


trishankatdatadog:
Where can we discuss this publicly to see whether it would break things for anyone upstream?

mayeut:
I guess the best place would be https://discuss.python.org/c/packaging/14

trishankatdatadog:
Would you mind starting the discussion, @mayeut ?

auvipy:
I think this can be closed now?

PR #440, opened 2020-01-18
Image versioning: Tag each docker build with a unique id, and push that to quay.io

### LINKED ISSUE #111: Introduce image versioning
It would be good to have images with a version rather than always overwriting the `latest` version.

Thus, one could use explicit versions.

My proposal would be to use the date as a version, e.g.
- `pypa/manylinux1_x86_64:20170502`

joerick:
Over at cibuildwheel we've been talking about how we might get some determinism in the build process (https://github.com/joerick/cibuildwheel/issues/239), and the manylinux image is one of those things we'd like to lock down.

This PR introduces a docker tag, alongside `latest`, of the format `YYYY-MM-dd-gitsha`, for example `2020-01-18-dea2616`. Each build is tagged, and the tag is pushed to quay.io.

I have tested this on my fork, with Travis building and uploading - see https://quay.io/repository/joerick/manylinux2010_x86_64?tab=tags for how these end up in quay.io.

Thanks!

Fixes #111

### DIFF
--- a/docker/deploy.sh
+++ b/docker/deploy.sh
@@ -1,5 +1,9 @@
 #!/bin/bash
-docker login -u $QUAY_USERNAME -p $QUAY_PASSWORD quay.io
 tag="quay.io/pypa/manylinux2010_$PLATFORM"
+build_id=$(git show -s --format=%cd-%h --date=short $TRAVIS_COMMIT)
+
+docker login -u $QUAY_USERNAME -p $QUAY_PASSWORD quay.io
+docker tag ${tag}:${TRAVIS_COMMIT} ${tag}:${build_id}
 docker tag ${tag}:${TRAVIS_COMMIT} ${tag}:latest
+docker push ${tag}:${build_id}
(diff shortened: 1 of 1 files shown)


joerick:
Any thoughts, @mayeut ?

mayeut:
@joerick,
I'm going to merge this but there might be a need to define a retention strategy in order not to store too much over the years.
Given it's been less than 100GB/year and that quay.io states an unlimited storage in its FAQ, that seems ok for now but nevertheless, we should think about that.

Can you create similar PRs for manylinux1 & manylinux2014 branches ?
Thanks

joerick:
> Can you create similar PRs for manylinux1 & manylinux2014 branches ?

Done, see #470 and #471.

> there might be a need to define a retention strategy in order not to store too much over the years.

Hm, yes. If quay.io could provide download stats, they could be used to delete images that weren't being used any more. I can't see that anywhere currently...

joerick:
Thanks you @mayeut !!

PR #497, opened 2020-03-19
use ensurepip, other small cleanups

mattip:
- document quay.io/organization/pypa for easy reference
- remove 3.4 from the documents
- replace get-pip with ensurepip
- add a hash for wheel
- pin python2.7 requirements since setuptools 46 doesn't support it xref gh-491)

### DIFF
--- a/docker/build_scripts/build_utils.sh
+++ b/docker/build_scripts/build_utils.sh
@@ -54,12 +54,12 @@ function do_cpython_build {
     if [ -e ${prefix}/bin/python3 ]; then
         ln -s python3 ${prefix}/bin/python
     fi
-    ${prefix}/bin/python get-pip.py
+    ${prefix}/bin/python -mensurepip
 
     if [ -e ${prefix}/bin/pip3 ] && [ ! -e ${prefix}/bin/pip ]; then
         ln -s pip3 ${prefix}/bin/pip
     fi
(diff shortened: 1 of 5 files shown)


docker/build_scripts/build_utils.sh
```diff
@@ -54,14 +54,18 @@ function do_cpython_build {
     if [ -e ${prefix}/bin/python3 ]; then
         ln -s python3 ${prefix}/bin/python
     fi
-    ${prefix}/bin/python get-pip.py
+    ${prefix}/bin/python -mensurepip
```
auvipy: -m ensurepip?

trishankatdatadog: Good catch! Was this PR tested?

mattip: why? Both `-mensurepip` and `-m ensurepip` work.

trishankatdatadog: I see. Yes, but the second one is more readable.

mattip: changing, thanks

mattip:
What is the proper git incantation to merge this to manylinux2014? Do we merge the branch into manylinux2014 or cherry-pick the commits?

mayeut:
@mattip, that would be cherry-picking the commits.

PR #499, opened 2020-03-20
[manylinux2010] Update CPython 3.7 from 3.7.6 to 3.7.7

manselmi:
Changelog: https://docs.python.org/release/3.7.7/whatsnew/changelog.html#python-3-7-7-final


mayeut:
thanks

PR #517, opened 2020-03-28
[manylinux1] Update OpenSSL from 1.0.2u to 1.1.1e

### LINKED ISSUE #429: Update from OpenSSL 1.0.2? to OpenSSL 1.1.1? (manylinux1 / manylinux2010)
OpenSSL 1.0.2 has reached end-of-life in December 2019: https://mta.openssl.org/pipermail/openssl-announce/2019-December/000164.html
Keeping up with security updates requires to update to the 1.1.1 series (LTS). This requires [dropping CPython 3.4](https://github.com/pypa/manylinux/issues/427)

mayeut:
Fix #429

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -3,5 +3,5 @@
 # Stop at any error, show all commands
 set -ex
 
-docker/build_scripts/prefetch.sh openssl curl
+docker/build_scripts/prefetch.sh perl openssl curl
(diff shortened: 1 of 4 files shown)


jamadden:
In https://bugs.python.org/issue40018 it's reported that the CPython ssl module is broken with OpenSSL 1.1.1e. This may cause problems for packages which attempt to use this module in the manylinux environment.

PR #545, opened 2020-04-18
Apply patch in patchelf to avoid zeroing SONAME

lkollar:
This patch is a copy of the upstream PR
https://github.com/NixOS/patchelf/pull/171. This fixes a bug in patchelf
which causes an SONAME entry filled with 'X' values.

This patch in `patchelf`fixes https://github.com/pypa/auditwheel/issues/159 until the patch is applied upstream.


myselfhimself:
THANK YOU!!

PR #549, opened 2020-04-24
[manylinux2014] Apply patch in patchelf to avoid zeroing SONAME

lkollar:
This patch is a copy of the upstream PR
https://github.com/NixOS/patchelf/pull/171. This fixes a bug in patchelf
which causes an SONAME entry filled with 'X' values.

Cherry picked https://github.com/pypa/manylinux/commit/052f2ffa38f812608bd549efe8ed1a441e2dd223.


lkollar:
The s390x build is failing again.

mayeut:
I'll restart the s390x build once the repo is fixed.

PR #562, opened 2020-05-01
Build i686 glibc

### LINKED ISSUE #561: Cannot install glibc-devel.i686 in manylinux2010_x86_64
```
# yum install glibc-devel.i686
...
Error: Package: glibc-2.12-1.212.el6_10.3.i686 (updates)
           Requires: glibc-common = 2.12-1.212.el6_10.3
           Installed: glibc-common-2.12-1.212.1.el6.x86_64 (@/glibc-common-2.12-1.212.1.el6.x86_64)
```

This must be because centos-6-no-vsyscall builds only 64-bit glibc. While I understand that the primary purpose of manylinux images is to build python wheels for specific architectures, it is still nice to have the base system fully functioning.

What do you think about building and installing a 32-bit glibc like this https://github.com/mephi42/manylinux/commit/82a54a74173fd7c3a3d4ffa2044b257d9cdbcdfb?

mephi42:
Fixes #561: this change fixes installing i686 packages into
manylinux2010_x86_64 image.

Also, it reworks Travis setup slightly, since x86_64 and i686 glibc
builds should run in different stages in order to avoid timeouts.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,10 +26,9 @@ env:
 _manylinux-build: &manylinux-build
       stage: "Build manylinux images"
       before_install:
-        # Load cached docker images
-        - if [[ -d $HOME/docker ]]; then ls $HOME/docker/*.tar.gz | xargs -I {file} sh -c "zcat {file} | docker load"; fi
+        - travis-ci/load_cached_images.sh combined
       script:
-        - PLATFORM=$PLATFORM TRAVIS_COMMIT=$TRAVIS_COMMIT ./build.sh
+        - PLATFORM=$PLATFORM TRAVIS_COMMIT=$TRAVIS_COMMIT ./build.sh glibc_skip
       deploy:
         provider: script
         script: docker/deploy.sh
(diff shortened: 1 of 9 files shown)


docker/glibc/Dockerfile
```diff
@@ -1,13 +1,33 @@
-FROM centos:6 as centos-with-vsyscall
+FROM i386/centos:6 as centos-with-vsyscall32
```
auvipy, line 1: why centos 6?

mephi42: This is for manylinux2010 - I'm using the same image here as the native manylinux2010 i686 build (i.e. `docker/Dockerfile-i686`).

mephi42:
The failure is due to a timeout. Most likely the existing timeout is tailored to building glibc once, but with this change it's done twice. Is it possible to increase the timeout?

mephi42:
Not sure whether Travis caches images, but in case it does, simply retriggering the build will help. I've rebased the commit - let's see if the new build passes (of course, if cache is cleared, then the next builds will fail - so it's probably better to increase the timeout anyway).

docker/glibc/Dockerfile
auvipy: ok got it

mephi42:
I think the good alternative to increasing timeouts is to split i686 glibc build into a separate stage. The updated patch aims to do that - I think I messed up setting `$PLATFORM` this time, so I'll push another update shortly.

mephi42:
Ok, so glibc builds don't timeout anymore, but their cached results are not picked up by the final manylinux build. I didn't find it documented explicitly anywhere, but some posts mention that travis cache is shared only between stages with identical `env`, which is not the case here. I will give sharing `env` a shot.

.travis.yml
```diff
@@ -37,19 +36,28 @@ _manylinux-build: &manylinux-build
...
+jobs:
+  include:
+    - <<: *patch-glibc
+      stage: "Build patched x86_64 glibc for x86_64 image"
+      script:
+        - PLATFORM=$PLATFORM docker/glibc/build.sh 32
+    - <<: *patch-glibc
+      stage: "Build patched i686 glibc for x86_64 image"
+      script:
+        - PLATFORM=$PLATFORM docker/glibc/build.sh 64
```
mayeut, line 56: What happens if `stage` is the same ? Would it get built in parallel thus speeding-up the build ? Does it interfere with caching ?

mephi42: Unfortunately, parallel build would indeed interfere with caching here. The problem is that even if you build a single target inside a Dockerfile, everything preceding it gets built, not just its dependencies. This makes 64-bit glibc effectively depend on the 32-bit one.

mephi42:
I checked Travis again, and seems like there are two things that can be fixed relatively easily:

- glibc Dockerfile may be split, which would allow building i686 and x86_64 concurrently.
- A lot of time is spent loading unnecessary docker images - since it is known in advance which ones are needed, it's better to load only these.

I'll push another commit with these changes.

mephi42:
At a closer look, parallel i686 / x86_64 glibc build is too bold of a goal, since even if they don't step on each other's toes, travis will refuse to "merge" their caches. I still split the Dockerfile, since it looks nicer this way.

I've also added finer-grained cache lookups and updates - this will hopefully speed things up.

Let's see how it goes.

mephi42:
@mayeut is there anything else I need to fix in order to get this change integrated?

mayeut:
@mephi42,

Thanks for the update. Looks good. Too bad about the // build but at least it gets cached properly.

I'm doing a last local check then I'll rebase the PR on master (same as my local check, please don't rebase) then I'll merge.

PR #565, opened 2020-05-03
[manylinux2010] Use devtoolset-8 on i686

### LINKED ISSUE #409: The compiler used in x86_64 is different than the one used in i686
PyPy started using the manylinux2010 docker images to build on x86. The compilers used in x86_64 and i686 are [different](https://github.com/pypa/manylinux/blob/master/docker/build_scripts/build.sh#L58), one is installed with `devtoolset-8` and the other with `devtoolset-7`. The devtoolset-8 compiler does not pass a test added in PyPy [2e8c4536e4](https://bitbucket.org/pypy/pypy/commits/2e8c4536e416b0c58ba0b1c3ca9e8eb702f46c48). The test can be described as testing  overflow behaviour when multiplying two ints with `-O3`: somehow **different** input to a complicated function yields the **same** output.

It seems to me that the two images should be using the same compiler, and I would prefer they both use devtoolset-7.

### LINKED ISSUE #543: GCC 7.4 or 8.3 on i686?
Hi,

I was just wondering if there is a chance to update to GCC 7.4, 7.5, GCC 8 (as on x86) or downgrade to 7.3.0 in the i686 docker images?

GCC 7.3.1 ICE's out on a popular C++ library: https://github.com/mpark/variant/issues/43 with this bug: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=84785

Re.: #355 cc @veblush

Thanks,
Axel

mayeut:
This PR will bring on i686 a gcc toolchain that's the same as on x86_64.

devtoolset-8 on i686 is rebuilt on `copr` infrastructure in the project [mayeut/devtoolset-8](https://copr.fedorainfracloud.org/coprs/mayeut/devtoolset-8/)

Before this can be merged, I'd like for @pypa/manylinux maintainers to request write access to the `copr` repo so that I'm not the only one with access.

Fix #543 
Fix #409

### DIFF
--- a/docker/build_scripts/build.sh
+++ b/docker/build_scripts/build.sh
@@ -52,32 +52,16 @@ esac
 yum -y update
 
 # EPEL support (for cmake28 & yasm)
-yum -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-6.noarch.rpm
+yum -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-6.noarch.rpm curl
 
 case $AUDITWHEEL_ARCH in
 x86_64)
-  # Install devtoolset-8
+  # Install centos-release-scl to get devtoolset-8
   yum -y install centos-release-scl
-  yum -y install \
-    devtoolset-8-binutils \
-    devtoolset-8-gcc \
-    devtoolset-8-gcc-c++ \
-    devtoolset-8-gcc-gfortran \
   ;;
 i686)
-  # Install devtoolset-7 (binutils, gcc, gcc-c++, gcc-gfortran)
-  devtoolset7s=(
-    "devtoolset-7-runtime-7.1-4.el6.i686.rpm"
-    "devtoolset-7-binutils-2.28-11.el6.i686.rpm"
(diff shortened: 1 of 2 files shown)


lkollar:
@mayeut I requested builder and admin access.

mayeut:
Thanks @lkollar, permissions granted.

docker/build_scripts/build.sh
```diff
@@ -87,6 +71,10 @@ yum -y install \
     bison \
     bzip2 \
     cmake28 \
+    devtoolset-8-binutils \
```
lkollar, line 74: What about the other packages which used to be installed, like `runtime`, `libstdc++` and `libquadmath-devel`?

mayeut: They are dependencies of these ones and are still installed.

tylerjereddy:
This seems to be having some effect on the scientific Python ecosystem--namely usage of our "standard" OpenBLAS binaries that were linked with a specific version of libgfortran is compromised. It is looking for an older runtime version of this lib than is now available in the 32-bit Docker image.

PR #627, opened 2020-06-09
Add CPython 3.9.0b3

### LINKED ISSUE #482: Support Python 3.9
rc1 was included for 3.8 (#313) not long before final. The current pre-release is 3.9.0a3. I would find it useful to have an alpha or beta personally but I don't know if that's ok generally.

ijl:
https://www.python.org/downloads/release/python-390b3/
https://www.python.org/dev/peps/pep-0596/

Fixes #482.


ijl:
Is there a reason?

lkollar:
b2 contains a regression so we should wait for b3: https://twitter.com/llanga/status/1270305329453170688

ijl:
Ok, it's building b3 now.

PR #738, opened 2020-08-20
Manylinux2014_aarch64: patch patchelf to fix alignment issues on Arm64

geoffreyblake:
Bug reports for Manylinux2014_aarch64 wheels show that they contain bad alignment for Arm distros that use 64k pagesize.  This updates patchelf 0.11 to include recent patches that force Arm64 page alignment to 64k regardless of the base system used for building.

Solves Issue: https://github.com/pypa/manylinux/issues/735


geoffreyblake:
Fat fingered it.

Issue #735, opened 2020-08-19
Inconsistent page-size on arm64 


ianw:
Hello,

tl;dr Debuntu has a 4k page-size and CentOS 7/8 has a 64k page size, so aarch64 manylinux wheels built on the former fail on the latter with alignment issues.

There seem to have been a number of places that have noticed this

* libffi build a wheel that caused problems for CI testing on pyca/cryptography : https://foss.heptapod.net/pypy/cffi/-/issues/468
* there is this numpy issue : https://github.com/numpy/numpy/issues/16677

and a few other issues pop up too.

Now I understand that to the letter of PEP599 that the build system should be CentOS 7; but in this case the page-size relates to the host system, so running the centos7 manylinux2014_aarch64 container is going to give you a different page-size depending on if you run it on Debuntu or CentOS (or of course, some other distro).

This is compounded by (AFAIK) travisCI only providing Ubuntu images (https://docs.travis-ci.com/user/reference/overview/ ?) making it difficult for people to get a 64k environment.

Individual builds can work around this with a linker flag to set 64k-alignment (e.g.https://github.com/pyca/bcrypt/commit/8d35d8acb82b4600a367e692d1231a705552900c) .  But that has to be ported into each build script and most people will never realise.

I wonder what solutions people might have for this?  Some I can think of

* raise a bug against devtoolset-9 to update the default alignment to 64k
* rebuilding devtoolset-9 with such a change and installing in the builder container
* some sort of hacky wrapper script around cc/ld to pass flags/linker-script/??
* some way to fail with a sensible error message; setuptools hack or something?
* redefine wheels to also publish the page-size they are built for (sounds painful, not backwards compatible and afaik, the only benefit is slightly smaller binaries in the 4k case ... if this could even be done with the plumbing required in pip etc).

It would be good to come up with something by default, because doing the obvious thing of grabbing the manylinux2014_aarch64 container and building wheels on debuntu (i.e. travis practically) creates wheels that don't actually work on the reference platform; so that seems suboptimal.

AGSaidi:
Atleast for gold this was fixed in the [linker to default to 64KB](https://sourceware.org/legacy-ml/binutils-cvs/2015-10/msg00050.html)

ianw:
@AGSaidi yes I tend to agree having played a bit more now.

I have a native Ubuntu Bionic arm64 host here, and it seems to create binaries 64k aligned both natively and also when building inside a manylinux2014_aarch container.  I'm basing this on 

```
[...]

PR #739, opened 2020-08-20
[manylinux2014] arm64 patchelf alignment fix

geoffreyblake:
Bug reports for Manylinux2014_aarch64 wheels show that they contain bad alignment for Arm distros that use 64k pagesize. This updates patchelf 0.11 to include recent patches that force Arm64 page alignment to 64k regardless of the base system used for building.

Tested locally with the cffi project, resulting wheel and binary after performing audit wheel uses 64k alignment.  Tested wheel on Centos8 on Arm64 and imported cffi with no issue.

Solves Issue: #735

### DIFF
--- a/docker/build_scripts/build_utils.sh
+++ b/docker/build_scripts/build_utils.sh
@@ -209,6 +209,7 @@ function build_patchelf {
     curl -fsSL -o patchelf.tar.gz https://github.com/NixOS/patchelf/archive/$patchelf_version.tar.gz
     check_sha256sum patchelf.tar.gz $patchelf_hash
     tar -xzf patchelf.tar.gz
-    (cd patchelf-$patchelf_version && ./bootstrap.sh && do_standard_install)
+    # Apply aarch64 alignment fix patch until version >0.11 of patchelf is released
+    (cd patchelf-$patchelf_version && patch -p1 -i "$src_dir"/patches/patchelf-fix-default-aarch64-alignment.patch && ./bootstrap.sh && do_standard_install)
     rm -rf patchelf.tar.gz patchelf-$patchelf_version
 }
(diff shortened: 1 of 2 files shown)


geoffreyblake:
Huh, Travis is failing on a curl call: 
`curl: (7) Failed to connect to 2a04:4e42:50::223: Cannot assign requested address`

Seems to be happening in some of the other PRs here too.  Additionally the PPC builds are being detected as stalled, and that seems to be happening for the past week or so:

```
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received
The build has been terminated
```

geoffreyblake:
@reaperhulk , @mattip , cherry-picking worked with just a minor merge issue. The resulting binary appears to be generating the proper alignments. Re-building the container to do a full test on a wheel.

geoffreyblake:
Internal testing on the pyca/cryptography shows it is generating binaries with the proper alignments on Arm.  Still seeing ppcle failing with timeouts, though all other builds are working.

mattip:
The goal of this PR is to finx the ARM64 images, so maybe it can just go in as-is, without the ppc64le builds. They have been failing for the past week.

reaperhulk:
Much easier to review! Could you add some text to make it clear that this patch can be backed out when patchelf is updated? Just in case someone else ends up doing that work. Otherwise this LGTM (although I am not a manylinux contributor so take my opinion with whatever sized grain of salt you desire)

geoffreyblake:
Ok, comments added.  I posted a question to the TravisCI mailing list about this PPC issue, seems odd it happens during a yum install.  This fix is primarily applicable to Arm64, which has multiple page sizes in use in the wild.

geoffreyblake:
@mattip, anything else needed for this? I do not believe that the PPCLE build is necessary for a fix related to Arm64.

mattip:
@geoffreyblake you need to ping someone who can merge.

reaperhulk:
@auvipy might be a good candidate here 😄

PR #740, opened 2020-08-28
Upgrade patchelf to 0.12

radarhere:
See #735


alex:
I'd think you need to revert the patching portions of https://github.com/pypa/manylinux/pull/739 as well.

radarhere:
Yes and no - that change was made on the manylinux2014 branch, and this PR is for the master branch. Apologies for not understanding the branch structure here - is this PR unhelpful?

radarhere:
For the record, the reversion has now been done in #741

mattip:
This PR does not really address #735 since that issue is arm64-specific. There may be other reasons to update patchelf, so I think it is worthwhile. One for the manylinux1 branch would bring all the branches into sync.

mayeut:
Thanks @radarhere. I also created a PR for manylinux1 to bring all the branches into sync.

PR #794, opened 2020-10-05
[manylinux1] CPython 3.9.0rc2 → 3.9.0

manselmi:
Changelog: https://docs.python.org/3.9/whatsnew/changelog.html


henryiii:
Shouldn't the readme (in the master branch) be updated to mention 3.9 now too?

PR #879, opened 2020-12-12
[manylinux2014] Rework scripts to be more docker cache friendly

mayeut:
This won't change travis-ci build time (at least for now) but allows maintainers/contributors to benefit from docker caching for local developments (faster builds at the expense of higher disk usage)

Round 1

Use docker multi-stage build to build the manylinux image in 3 steps:

1. runtime_base: this image has all the necessary bits that are common to the build_base image and the manylinux image. This includes system packages required for runtime, the gcc toolchain & some basic tools.

2. build_base: this image builds all the remaining tools required for the manylinux image. Any tool requiring specific system development package not required in the manylinux image shall be built from this image. For tools that used to be installed in /usr/local, they are staged in /manylinux-rootfs for an easy copy in the third stage.

3. manylinux: This image uses the runtime_base image as a base image. Tools built in the 2nd step are copied directly in the final filesystem. A finalization script is then run to install the remaining python tools/dependencies and run checks.

This PR builds on top of #876


mattip:
Is this ready for final review?

mayeut:
This one is ready for review. I merged the previous PR without any reviews in order to get things going forward even though I don't like to do this without reviews...
It's only a part of multiple PR that are yet to come to get in a state I find acceptable for a single branch, docker cache friendly state.
I won't wait for more than a week for further reviews. If there are none, I will assume this is acceptable.

mattip:
#965 is next?

mayeut:
@mattip, yes, I'll merge everything up to building on GHA which will take some time for aarch64/s390x/ppc64le

mattip:
thanks!

Issue #877, opened 2020-12-12
Redesign scripts to build all images from a single branch


mayeut:
As proposed in https://github.com/pypa/manylinux/issues/542#issue-598574781, this issue is created to propose and track progress of a rewrite of `manylinux` build scripts to build all images from a single branch.

With new `manylinux` policies allowed by PEP600, we'll have more and more images. Creating a branch per policy will cause issues for maintainability and consistency between images (or even README).

`manylinux1` is out-of-scope for the rewrite for now (c.f. https://github.com/pypa/manylinux/issues/542#issuecomment-720674410 for rational)

Here are the steps I propose, feel free to challenge those.
PRs referenced here will build one on top of the other requiring them to be reviewed sequentially.

- [x]  Rework the `manylinux2014` branch to allow for multiple policies
Rational for choosing this as the base branch of the rework: `manylinux2014` branch builds on all platforms which allows to see issues that would otherwise not be seen until a later point.

  - [x] Use graviton2 for arm64 build (#875)
  - [x] Use a single Dockerfile for all images (#876)
  - [x] Rework scripts to be docker cache friendly (mostly for maintainers doing local builds)
    - [x] Round 1 (#879)
    - [x] Round 2 (#965) 
    - [x] Round 3 (#966)
    - [x] Round 4 (#967)
    - [x] Round 5 (#968)
    - [x] Round 6 (#969)
    - [x] Round 7 (#970)
    - [x] Round 8 (#971)
    - [x] Round 9 (#972)
    - [x] Round 10 (#973)
    - [x] Round 11 (#974)
    - [x] Round 12 (#975)
    - [x] Round 13 (#976)
    - [x] Round 14 (#977)
    - [x] Round 15 (#978)
- [x] Merge `manylinux2014` in `master`
At this point, we'll be ready to add new images for newer PEP600 policies
  - [x] Update `glibc` on `manylinux2010` (#874)
  - [x] Publish `quay.io/pypa/manylinux2010_centos-6-no-vsyscall` (#878)
  - [x] Add `manylinux2010` build support in `manylinux2014` branch (#979)
  - [x] Merge in `master` (#993)

- [x] Add `manylinux_2_24` images (#999)
[...]

PR #918, opened 2020-12-24
[manylinux2010] CPython 3.8.6 → 3.8.7

manselmi:
Changelog: https://docs.python.org/3.8/whatsnew/changelog.html#python-3-8-7-final


mayeut:
Thanks for the PR @manselmi. I'm looking at a way to get these automated. Although I'd rather have #877 done before doing that.

manselmi:
> Thanks for the PR @manselmi. I'm looking at a way to get these automated. Although I'd rather have #877 done before doing that.

Sounds great!

PR #941, opened 2021-01-06
feat: add pypa/build

### LINKED ISSUE #940: Include pypa/build
The manylinux images should, I think, include https://github.com/pypa/build by default, as that's the PyPA wheel builder.

henryiii:
This adds pypa/build to the requirements file. CC @FFY00 

Closes #940.

### DIFF
--- a/docker/build_scripts/requirements.txt
+++ b/docker/build_scripts/requirements.txt
@@ -15,3 +15,34 @@ setuptools==50.3.2 ; python_version=='3.5' \
 setuptools==53.1.0 ; python_version>='3.6' \
      --hash=sha256:487f7a3e5040445b687f1eadcebcbbd11e83da95cad5e755304e38a3b7a93207 \
      --hash=sha256:8181c4625cd58fa7817761e0fa60f6fa3adc4c437f5c25377b1aefa861ca15a2
+build==0.3.0 \
+    --hash=sha256:0eb95b2c8d770d7c4c9b92c68c227c350bbf65f3ec83551ace9097c18cc15fdd \
+    --hash=sha256:75bc5676b1a014fb996dc96914428b197d174364392a46778200d5e115ffe76c
+packaging==20.9 \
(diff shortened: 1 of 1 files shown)


mayeut:
Unfortunately, this doesn't pass CI tests because it's expecting the sha sums of all wheels (i.e. not only top-level).
I'm already finding `dependabot` quite not fitted with the low number of dependencies we currently have and, given the number of dependencies that `pypa/build` will require, I'm not keen on adding this just yet.
I'm thinking of rewriting python dependencies update a bit more like what's being done over at `cibuildwheel` to handle the multiple versions of python in a cleaner way.
There's also the fact that, at the moment, we require 3 branches to be updated. There's #877, #876, #879 but it seems we're short on `manylinux` maintainers review time. I'd also be glad to see other contributors/users feedback if that has  any impact for them (I expect no impact for most users, only people rebuilding the image themselves with some patches will be highly impacted by the rewrite).

henryiii:
I believe it's 3-6 dependencies, just `packaging`, `toml`, and `pep517` for the most recent Python, though older versions of Python 3 also need `importlib-metadata`, and Python 2 needs `virtualenv` and `typing`. I believe these are technically almost all present in vendored form in pip, actually. ;)

But if you hope to refactoring the system a bit, this doesn't have to go in instantly. Targeting three branches is irritating (and might get worse with the new `manylinux_2_*` naming)?, would love to see that improved, but I'm happy to add the pins and target the branches if needed.

henryiii:
A pip-compile based workflow with multiple requirements.txt's and a single requirements.in would be much nicer :)

mayeut:
Thanks @henryiii 

Yes, the requirements.txt is a mess, mostly because of how dependabot works and workaround some issues with it. This will be reworked at a later date with a full dependency update workflow bypassing completely dependabot issues (like what's done in cibuildwheel).

PR #950, opened 2021-01-14
Add Python 3.10a4 Support

### LINKED ISSUE #949: Adding support for 3.10 pre-release to manylinux images
Currently Python 3.10 is in Alpha 4 (3.10a4). Any chance of including 3.10-dev, or the latest candidates in images? This would enable projects I work on to get ahead of the release for 3.10 and avoid delays in delivering wheels for customers using the latest python versions.

Thanks!

crwilcox:
This fixes #949 by adding the alpha candidate for 3.10, along with necessary keys.


crwilcox:
@mayeut do you have an opinion on this? I think it would be useful to be able to use the docker images to build against the upcoming release (https://www.python.org/downloads/release/python-3100a4/)

Also note, alpha 5 is expected *soon* so I can rev this before we merge also.

mayeut:
@crwilcox,
We usually wait for a beta release for new python because the C API/ABI is allowed to change in alpha stage which might lead to all sorts of bugs if someone deploys something built with an alpha version on PyPI.

crwilcox:
@mayeut that is reasonable. If this causes burden to keep up to date I can definitely understand, though I would find value in this since I am hoping to flight these builds and hopefully let CPython dev know if something goes awry. That said, waiting until beta isn't a huge deal :)

Issue #428, opened 2020-01-13
Drop support for CPython 2.7 (manylinux2010)


mayeut:
CPython 2.7 has reached end-of-life in January 2020

lkollar:
This could impact many users, but I wonder if we could tag the last image still shipping with 2.7 and keep it around as a fallback for users who still need it?

rdb:
CPython 2.7 won't be EOL until April 2020:
https://www.python.org/psf/press-release/pr20191220/

lkollar:
We could create and tag the last image once Python 2.7.18 is released during PyCon just so that users have the last version available.

mayeut:
Since end of january, all images are tagged so tagging => OK
Once Python 2.7.18 is released, we'll update to get the proper tag then drop python 2.7.

hugovk:
Python 2.7.18 has been released in April! 🚀

pradyunsg:
Whenever we're ready. ;)

mayeut:
@pradyunsg,

The question would be "whenever package maintainers are ready".
After digging around some stats for different subjects around `manylinux`, I'm not so sure they are...

I'll write something here but please tell me if it's better suited for https://discuss.python.org/c/packaging.

I made some stats on all latest release of projects producing manylinux wheels, active (having at least 1 release in the past year) and for which at least 1 manylinux wheel has been downloaded over the past month. As with all stats, it might be biased by my process and/or interpretation.
I checked some metrics for projects having their latest release going back 12 months, 6 months, 3 months & 1 month.

Here are some interesting numbers regarding python versions:

  | 12 months | 6 months | 3 months | 1 month
-- | -- | -- | -- | --
number of projects with manylinux wheels | 2007 | 1543 | 1109 | 608
cp26 | 0,1% | 0,1% | 0,1% | 0,2%
cp27 | 26,1% | 22,4% | 22,3% | 20,9%
cp33 | 0,1% | 0,0% | 0,0% | 0,0%
cp34 | 5,0% | 2,5% | 2,3% | 2,8%
cp35 | 47,8% | 45,8% | 44,3% | 42,3%

For those still building for cp26/cp33, they are using custom images as tagging every manylinux image in quay.io occurred after those versions were dropped. For those building for cp34, they might be using a tagged manylinux image.

As for cp27, if all those package maintainers are using a single image workflow and still want to maintain cp27, this will lead to roughly 500 package build workflow that will need to be reworked.

[...]

PR #993, opened 2021-02-07
Merge manylinux2014 branch in master branch

### LINKED ISSUE #659: Add recent CMake and SWIG
It would be great if the docker images came with a recent cmake and swig pre-installed, since these are fairly common dependencies for Python packages that need to be build.

mayeut:
Part of #877 

CPython 2.7 will be dropped on manylinux2010 (#428)

Fix #659


rdb:
Why SWIG?  Is there a particular reason for including it?

mayeut:
c.f. #659

PR #998, opened 2021-02-13
Build ppc64le on travis-ci

mayeut:
Now that pypa has some credits on travis-ci again, move back ppc64le on travis-ci where builds are done natively.

- We need to run a rootless docker daemon on ppc64le due to travis-ci LXD configuration but this is better than the
6 hour build on GHA reaching the time limit now & then. We can still go back to QEMU GHA if things prove unstable.

- Disable parallel builds: Given the dockerfile uses multiple unrelated stages & that we're using buildx/buildkit which parallelize those stages, parallelizing builds at each stage does not improve build speed that much and puts too much RAM pressure leading to random failures in CI.

- Workaround GPG issue with ppc64le build: Seems there's an issue with GPG when using the rootless docker daemon,
the import statement and the verify statement must be in the same layer.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -30,14 +30,10 @@ jobs:
             platform: "i686"
           - policy: "manylinux2014"
             platform: "x86_64"
-          - policy: "manylinux2014"
-            platform: "ppc64le"
           - policy: "manylinux_2_24"
             platform: "i686"
           - policy: "manylinux_2_24"
             platform: "x86_64"
-          - policy: "manylinux_2_24"
-            platform: "ppc64le"
 
     env:
       POLICY: ${{ matrix.policy }}
@@ -47,12 +43,6 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
-
-      - name: Set up QEMU
-        if: matrix.platform == 'ppc64le'
-        uses: docker/setup-qemu-action@v1.0.1
-        with:
-          platforms: ppc64le
 
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v1
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,21 +28,23 @@ jobs:
       env: POLICY="manylinux2014" PLATFORM="aarch64"
     - arch: s390x
       env: POLICY="manylinux2014" PLATFORM="s390x"
-    #- arch: ppc64le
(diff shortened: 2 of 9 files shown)


mayeut:
https://travis-ci.community/t/ppc64le-docker-buildx-issues/11180

adr31 (commented):
configurar Docker Buildx

.github/workflows/build.yml
```diff
@@ -30,14 +30,10 @@ jobs:
             platform: "i686"
           - policy: "manylinux2014"
```
adr31, line 31: .configurar Docker Buildx

.travis.yml
```diff
@@ -28,16 +28,16 @@ jobs:
       env: POLICY="manylinux2014" PLATFORM="aarch64"
     - arch: s390x
       env: POLICY="manylinux2014" PLATFORM="s390x"
-    #- arch: ppc64le
-    #  env: POLICY="manylinux2014" PLATFORM="ppc64le"
+    - arch: ppc64le
+      env: POLICY="manylinux2014" PLATFORM="ppc64le"
     - arch: arm64-graviton2
```
adr31, line 33: configurar Docker Buildx

adr31 (approved):
configurar Docker Buildx

webknjaz:
> [travis-ci.community/t/ppc64le-docker-buildx-issues/11180](https://travis-ci.community/t/ppc64le-docker-buildx-issues/11180)

Have you considered using a rootless podman invocation instead?

mayeut:
Thanks for the suggestion @webknjaz 
It appears that podman does not support caching which would be a no-go IMHO given the limited resources we now have on travis-ci.
I might try it anyway just to learn some things.

webknjaz:
Yeah, that is indeed annoying, I can't wait for buildah to fix the caching.
The reason I suggested it was that I faced some problems with QEMU jobs I was inheriting from the pyca/infra project. And just changing to podman magically fixed things so I never dug into why this way happening.

qa8302:
https://workspace.google.com/intl/en_uk/security/
https://hiip.wpro.who.int/portal/default.aspx
  17-end-user-undertaking-form.pdf
<https://drive.google.com/file/d/1NB9dE8zgZ8kspN8AJROUqJPKCFb3bqf7/view?usp=drivesdk>


Sviatoslav Sydorenko ***@***.***> 於 2021年3月11日 週四 上午12:42 寫道：

Issue #1005, opened 2021-02-18
CVE-2021-3177


aws-taylor:
Hello ManyLinux team,

  Redhat just announced a new security vulnerability detected in python - https://access.redhat.com/security/cve/cve-2021-3177 - that appears to impact the interpreters shipped in the manylinux containers (or at least 2010). Given that manylinux deliberately ships libpython.a, it seems like teams that are using this container to build portable binary python .whl files may be at risk.  At this time, python 3.6 and 3.7 appear to be patched, so I think the fix is simply updating the minor version in the Dockerfile. It appears that 3.8 and 3.9 are not yet fixed.

The related issue https://bugs.python.org/issue42938 has a short test to reproduce:

```
/opt/python/cp35-cp35m/bin/python3 -c 'import sys; from ctypes import *; print(c_double.from_param(1e300));
Segmentation fault
```

This may also be a good reason to revisit the discussion in https://github.com/pypa/manylinux/issues/59 with regards to fortifying the python interpreters.

di:
Thanks! 3.8.8rc1 and 3.9.2rc1 will be promoted [on Monday March 1st](https://discuss.python.org/t/python-3-9-2rc1-and-3-8-8rc1-are-now-available-for-testing/7253), I think at that point we can bump all the versions in the Dockerfile and release new images.

mayeut:
@di, thanks for reminding the timeline for 3.8 & 3.9

@aws-taylor , thanks for the info

> Given that manylinux deliberately ships libpython.a, it seems like teams that are using this container to build portable binary python .whl files may be at risk.

manylinux does not ship libpython.a (c.f. #91 requesting for it to be added). So no, wheels are not affected, only the interpreters are.

aws-taylor:
> @di, thanks for reminding the timeline for 3.8 & 3.9
> 
> @aws-taylor , thanks for the info
> 
> > Given that manylinux deliberately ships libpython.a, it seems like teams that are using this container to build portable binary python .whl files may be at risk.
> 
> manylinux does not ship libpython.a (c.f. #91 requesting for it to be added). So no, wheels are not affected, only the interpreters are.

Great! Thanks for the clarification. I see now that it was being included via some docker volume/-v mounts that I was passing in my local test.

[...]

Issue #59, opened 2016-04-28
Enable hardening of binaries included in the wheels built for manylinux 


wallrj:
We package our Python software and all its dependencies  as a Debian package for easy installation and distribution.

We run `lintian` on the deb package files and recently it began issuing the following warning:

```
W: clusterhq-python-flocker: hardening-no-relro opt/flocker/lib/python2.7/site-packages/msgpack/_packer.so
W: clusterhq-python-flocker: hardening-no-relro opt/flocker/lib/python2.7/site-packages/msgpack/_unpacker.so
```

(Our ref https://clusterhq.atlassian.net/browse/FLOC-4383)

We think it's because we recently updated to `pip==8.1.1` which installs manylinux binary wheel files.
And the binaries in these wheels are not compiled with the hardening features that are required of binaries in Debian packages:

```
(venv)root@6dcaee731129:/# hardening-check /tmp/venv/lib/python2.7/site-packages/msgpack/*.so
/tmp/venv/lib/python2.7/site-packages/msgpack/_packer.so:
 Position Independent Executable: no, regular shared library (ignored)
 Stack protected: no, not found!
 Fortify Source functions: no, only unprotected functions found!
 Read-only relocations: no, not found!
 Immediate binding: no, not found!
/tmp/venv/lib/python2.7/site-packages/msgpack/_unpacker.so:
 Position Independent Executable: no, regular shared library (ignored)
 Stack protected: no, not found!
 Fortify Source functions: no, only unprotected functions found!
 Read-only relocations: no, not found!
 Immediate binding: no, not found!
```

Perhaps the manylinux build environment could set the necessary environment variables e.g:

```
dpkg-buildflags --export
export CFLAGS="-g -O2 -fstack-protector --param=ssp-buffer-size=4 -Wformat -Werror=format-security"
export CPPFLAGS="-D_FORTIFY_SOURCE=2"
export CXXFLAGS="-g -O2 -fstack-protector --param=ssp-buffer-size=4 -Wformat -Werror=format-security"
export FFLAGS="-g -O2 -fstack-protector --param=ssp-buffer-size=4"
export GCJFLAGS="-g -O2 -fstack-protector --param=ssp-buffer-size=4"
export LDFLAGS="-Wl,-Bsymbolic-functions -Wl,-z,relro"

```

Or use http://manpages.ubuntu.com/manpages/wily/man1/hardening-wrapper.1.html

Maybe related to #46

rmcgibbo:
[...]

PR #1009, opened 2021-02-21
Build hardening

mayeut:
- Use hardening for building all tools & libraries
This does not affect the wheels that are produced by end users as proposed in #59 but mitigates potential security issues in the tools used by manylinux images as mentioned in #1005
- Always update system packages in the final step
Since docker cache is used, system packages are not updated when cache is present. Always update them in the final step.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -21,18 +21,17 @@ ENV LD_LIBRARY_PATH=${LD_LIBRARY_PATH_ARG}
 ENV PATH=${PREPEND_PATH}${PATH}
 ENV PKG_CONFIG_PATH=/usr/local/lib/pkgconfig
 
-# first fixup mirrors, keep the script around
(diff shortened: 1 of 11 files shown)


mayeut:
@trishankatdatadog, @aws-taylor,  @wallrj,

Any thoughts on this PR ?

trishankatdatadog:
Good idea. What about all the [other default flags](https://wiki.ubuntu.com/ToolChain/CompilerFlags#Default_Flags)?

mayeut:
All other flags I can think of for security purposes do not work on those old distros:
```
-fPIE / -Wl,-pie
-fstack-clash-protection
-fcf-protection
```

trishankatdatadog:
> All other flags I can think of for security purposes do not work on those old distros:

Understood. If so, let's link to that document I sent, and comment on why we turned on only some flags and not others?

PR #1017, opened 2021-02-28
Add default pip compatibility for CentOS 7/8 (rh-)python38

mayeut:


henryiii:
Wow, CentOS 8's default `python3` is Python 3.6 with pip 9...

mayeut:
yep...

https://bugzilla.redhat.com/show_bug.cgi?id=1821733

From what I understand (I may be wrong) they won't do anything for python 3.6 unless a paid customer asks for it.
Since RHEL now provides "Application Streams" with "up-to-date" user apps, I guess that `python38` supporting this is enough for them (and it's easier to install/use than `rh-python38` on CentOS 7 which requires installing `centos-release-scl-rh` before and explicitly enable it with `source /opt/rh/rh-python38/enable `)

PR #1018, opened 2021-02-28
[manylinux1] backport build

henryiii:
Followup to #941 for manylinux1.

### DIFF
--- a/docker/build_scripts/requirements.txt
+++ b/docker/build_scripts/requirements.txt
@@ -18,3 +18,79 @@ setuptools==50.3.2 ; python_version=='3.5' \
 setuptools==53.1.0; python_version>='3.6' \
      --hash=sha256:487f7a3e5040445b687f1eadcebcbbd11e83da95cad5e755304e38a3b7a93207 \
      --hash=sha256:8181c4625cd58fa7817761e0fa60f6fa3adc4c437f5c25377b1aefa861ca15a2
+build==0.3.0 \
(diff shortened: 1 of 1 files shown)


henryiii:
This would be _so_ much shorter if Python 2 was dropped. Why not drop Python 2 for manylinux1 as well? And drop Python 3.5 in May along with the other manylinuxes?

auvipy (approved):

mayeut:
@henryiii,

I'm not planning to do any more work on `manylinux1` except bumping versions for updates. No new enhancements, no removals. Just basic support with feature freeze while waiting for EOL. We'll see how usage evolves by the end of the year and I can revisit this position at a later date depending on that.
Given you did the work for `build`, merging this enhancement hopping it won't get messy with dependabot.

henryiii:
I would think that Python 2 removed from manylinux1, and Python 3.5 removed in May. By _not_ removing them, these are getting maintained _longer_ that if removed now. Someone using Python 2 is supposed to be using a pinned manylinux{^1} version already, why not require it across all manylinuxs? It would be much easier to drop Python 2 support in tools like cibuildwheel if all manylinuxs have dropped it. More importantly to me, Python 3.5 will "live" much longer if it's not removed when removed from the other manylinuxes in a few months. It's better to maintain less!

henryiii:
PS: If you need me to, I'm happy to do a removal.

PR #1049, opened 2021-03-22
sqlite3 headers

drewgilliam:
Previous versions of manylinux2010 included sqlite3 files that allowed downstream dockers to build against the recent version of sqlite3 included in the docker image.  Files include:
- headers `/usr/local/include/{sqlite3.h,sqlite3ext.h}`
- pkgconfig file `/usr/local/lib/pkgconfig/sqlite3.pc`
- libtool file `/usr/local/lib/libsqlite3.la`
- .so file `/usr/local/lib/libsqlite3.so`

PR #972 eliminated these files during cleanup.  This PR restores the above files (which have a relatively small footprint).

### DIFF
--- a/tests/ctest/CMakeLists.txt
+++ b/tests/ctest/CMakeLists.txt
@@ -4,7 +4,7 @@ include(CTest)
 
 # SQLite3 test Derived from CMake unit test for FindSQLite3
 # https://gitlab.kitware.com/cmake/cmake/-/tree/master/Tests/FindSQLite3/Test
-find_package(SQLite3 REQUIRED)
+find_package(SQLite3 3.34 REQUIRED)
 add_definitions(-DCMAKE_EXPECTED_SQLite3_VERSION="${SQLite3_VERSION}")
 add_executable(test_sqlite3 test_sqlite3.cpp)
 target_link_libraries(test_sqlite3 SQLite::SQLite3)
--- a/tests/ctest/CMakeLists.txt
+++ b/tests/ctest/CMakeLists.txt
@@ -1,0 +1,11 @@
+cmake_minimum_required(VERSION 3.14.7)
+project(manylinux_ctest)
+include(CTest)
+
+# SQLite3 test Derived from CMake unit test for FindSQLite3
+# https://gitlab.kitware.com/cmake/cmake/-/tree/master/Tests/FindSQLite3/Test
+find_package(SQLite3 REQUIRED)
+add_definitions(-DCMAKE_EXPECTED_SQLite3_VERSION="${SQLite3_VERSION}")
+add_executable(test_sqlite3 test_sqlite3.cpp)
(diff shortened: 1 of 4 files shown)


auvipy (approved):

drewgilliam:
@auvipy - thanks for the approval!  Not sure what to do about the failing Travis build, which appears to fail way before it gets to the SQLite install.  I know quay.io was down right when I submitted the PR, not sure if that mattered here.  Any suggestions?

mayeut:
Hello @drewgilliam,

Thanks for the PR. Indeed there was a regression here.
  
While waiting for https://github.com/nealef/clefos/issues/7 resolution (will fix the s390x failure in CI), would you be willing to also contribute a small test to ensure this does not happen again ?
All tests were moved in the `tests` folder. Something really small would do just fine, if possible, using all files that have been added back.

drewgilliam:
@mayeut - I added a simple ctest structure based on the CMake unit test for FindSQLite3
https://gitlab.kitware.com/cmake/cmake/-/tree/master/Tests/FindSQLite3/Test

This should test the intent of the PR, being able to compile & run against sqlite3.  The simple ctest system can also be easily extended with other tests in the future with similar tests.

Note I also added `sqlite3 --version` to the minimal test section of `run_tests.sh` which required keeping the `/usr/local/bin/sqlite3` executable around.  FYI, it appears there may be a competing version of sqlite3 in some docker images (e.g., `/usr/bin/sqlite3` with a different version) though `/usr/local/bin/sqlite3` should be favored by the path.

tests/ctest/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+cmake_minimum_required(VERSION 3.14.7)
+project(manylinux_ctest)
+include(CTest)
+
+# SQLite3 test Derived from CMake unit test for FindSQLite3
+# https://gitlab.kitware.com/cmake/cmake/-/tree/master/Tests/FindSQLite3/Test
+find_package(SQLite3 REQUIRED)
```
mayeut, line 7:
```suggestion
find_package(SQLite3 3.34 REQUIRED)
```

Check for 3.34+ to ensure the manylinux built version is found. This won't need to be updated each time sqlite is updated.

mayeut (changes requested):
>  FYI, it appears there may be a competing version of sqlite3 in some docker images (e.g., /usr/bin/sqlite3 with a different version) though /usr/local/bin/sqlite3 should be favored by the path.

This might also be the case for other tools & I'd like to address this potential issue at some point with a better check than just printing out the version (which only ensures that some version of the tool is in the PATH).
For now this will be enough given `/usr/local/bin` is before `/usr/bin` in the PATH.

mayeut (approved):

mayeut:
Thanks @drewgilliam

PR #1055, opened 2021-03-30
refactor: generate requirements files

### LINKED ISSUE #1019: pip-compile based requirements.txt
I've started thinking about a pip-compile based requirements generation; I've started by writing this `noxfile.py`:

```python
import nox

@nox.session(python=["2.7", "3.5", "3.6", "3.7", "3.8", "3.9"])
def compile(session):
    session.install("pip-tools")
    session.chdir("docker/build_scripts")
    session.run("pip-compile", "--generate-hashes", "requirements.in", "--output-file", f"requirements{session.python}.txt")
```

And this `requirements.in`:

```
pip
setuptools
wheel
build
```

However, pip and setuptools are missing from the output. Maybe a known issue with pip-compile? (since they are already installed) Don't know, so just dumping the current status here.

henryiii:
Closes #1019, based on the similar structure in cibuildwheel. Suggested by @mayut to improve maintainability.

To run locally, either install all versions of Python or use docker, two examples below:

```console
docker run --rm -v $PWD:/nox -t thekevjames/nox nox -f /nox/noxfile.py -s compile tools
docker run --rm -v $PWD:/nox -t quay.io/pypa/manylinux2010_x86_64:latest pipx run nox -f /nox/noxfile.py -s compile tools
```


docker/build_scripts/finalize.sh
```diff
@@ -20,9 +20,10 @@ for PREFIX in $(find /opt/_internal/ -mindepth 1 -maxdepth 1 -name 'cpython*');
 	if [ -e ${PREFIX}/bin/pip3 ] && [ ! -e ${PREFIX}/bin/pip ]; then
 		ln -s pip3 ${PREFIX}/bin/pip
 	fi
+    PY_VER=$(${PREFIX}/bin/python -c "import sys; print('.'.join(str(v) for v in sys.version_info[:2]))")
```
mayeut, line 23:
```suggestion
	PY_VER=$(${PREFIX}/bin/python -c "import sys; print('.'.join(str(v) for v in sys.version_info[:2]))")
```
minor style issue to be consistent with the rest of the file.

mayeut (changes requested):
Thanks @henryiii,  this looks good overall.
One minor issue is that we'll lose what's been bumped in the PR description and the change logs from dependabot but those minor details are really not showstoppers and, at least for what's been bumped, one can look at the commit itself rather than its description.

mayeut:
I might use this PR as a baseline to continue with the automated update thingy I worked on a couple months ago.
I also wanted to have a look at `nox` (though not for `manylinux`) so this might be the occasion.

.github/workflows/update-dependencies.yml
```diff
@@ -0,0 +1,28 @@
...
+  schedule:
+    - cron: '0 6 * * 1'  # "At 06:00 on Monday."
+
+jobs:
+  update-dependencies:
+    name: Update dependencies
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: excitedleigh/setup-nox@1.0.0
```
mayeut, line 14: I'm all for not "reinventing the wheel" but this doesn't seem very popular as of now which might raise some maintenance/availability issues. The repo is "active" and we can rework this later if this proves necessary.

henryiii: It's "official" (meaning it is listed on the nox main page), but it was broken for a long time. Originally, it forked the actions/setup-python script and modified it to run on each version of Python. This was not maintainable! About a month ago (my sense of time may be off :baby:), it was redesigned into a very simple action that uses the JS API and now should be trivial to maintain.

And, it's completely replaceable. It's identical to the following block of YAML:

```yaml
- uses: actions/setup-python@v2
  with:
    python-version: '3.6'
- uses: actions/setup-python@v2
  with:
    python-version: '3.7'
- uses: actions/setup-python@v2
  with:
    python-version: '3.8'
- uses: actions/setup-python@v2
  with:
    python-version: '3.9'
```

henryiii:
@mayeut Should I add docs? If you think this approach is better, let me know how you want it finished off.

.github/workflows/update-dependencies.yml
```diff
@@ -0,0 +1,28 @@
...
+    - name: "Run update: dependencies"
+      run: pipx run nox
+    - name: Create Pull Request
+      if: github.ref == 'refs/heads/master' && github.repository == 'pypa/manylinux'
+      uses: peter-evans/create-pull-request@v3
```
henryiii, line 19: FYI, `gh` could probably do this as well, see https://github.blog/2021-03-11-scripting-with-github-cli/#using-gh-in-github-actions

mayeut: Good to know.
I think I'll get a look at `gh` when updating the workflow to check more dependencies.
For now the simple `peter-evans/create-pull-request@v3` is enough I guess.

mayeut:
@henryiii,

I'll have another look once CPython 3.5 is dropped (and CPython 3.10 added)

mayeut:
Now that #1090 is merged in, I wonder if `nox` or `pipx` would be good additions in order to run nox directly in the manylinux image.

See also:
https://github.com/joerick/cibuildwheel/pull/661

henryiii:
I used #1090 to run nox in manylinux to get Python 3.10. :)

henryiii:
Note: if the nox action on GHA doesn't include pre-release Pythons, then we'll need one extra setup-python with 3.10.

henryiii:
PS: It's also still very much up to you which method you'd rather maintain. :)

mayeut:
> Note: if the nox action on GHA doesn't include pre-release Pythons, then we'll need one extra setup-python with 3.10.

Now that we talked about that both here & over at cibuildwheel, I opened #1095 with the addition of pipx and it would now makes sense, I think, to use one of the manylinux image to run the job (& in the doc).

For the CI job, I'd go with `manylinux_2_24_x86_64` which is the smallest one.

For the docs, it might be worth mentioning that any manylinux2010+ image can be used.

henryiii:
I don't think we want to run the job in the image, since it's probably faster to use the native actions over pulling a docker container, and you get into a chicken-and-egg problem if you add a new Python version.

(Yes, it is, which is why I used #1090 to run the 3.10 job in manylinux, manually adding nox - should be much simpler now)

mayeut:
> I don't think we want to run the job in the image, since it's probably faster to use the native actions over pulling a docker container, and you get into a chicken-and-egg problem if you add a new Python version.

Sorry, I missed the fact that the nox image wasn't used in the job. I'm understanding your previous Note correctly now.
I think still my comment is valid for the docs though.

For the chicken-and-egg problem, my best guess is that when adding a new python version, we'll just need to copy/paste the previous python version requirement file to bootstrap the whole process. The job would then just pick that up no ? It doesn't happen that often anyway for it to be really troublesome

.github/workflows/update-dependencies.yml
```diff
@@ -0,0 +1,28 @@
+name: Update dependencies
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 6 * * 1'  # "At 06:00 on Monday."
```
mayeut, line 6:
```suggestion
    - cron: '0 6 * * 1'  # "At 18:00 on Sunday."
```
This would let me have some time to review (not that much probably)/merge this before the cibuildwheel update which is still at 06:00 on Monday and thus will have a bit more tests run before the business week really starts.

README.rst
```diff
@@ -175,6 +175,15 @@ current (root) directory:
...
 Please note that the Docker build is using `buildx <https://github.com/docker/buildx>`_.
 
+Updating the requirements
+-------------------------
+
+The requirement files are pinned and controlled by pip-tools compile. To update
+the pins, run nox on a Linux system with all supported versions of Python included.
+For example, using a docker image:
+
+    $ docker run --rm -v $PWD:/nox -t thekevjames/nox nox -f /nox/noxfile.py -s compile tools
```
mayeut, line 185:
```suggestion
    $ docker run --rm -v $PWD:/nox -t quay.io/pypa/manylinux2010_x86_64:latest pipx run nox -f /nox/noxfile.py -s compile tools 
```

noxfile.py
```diff
@@ -0,0 +1,31 @@
...
+    session.run(
+        "pip-compile",
+        "--generate-hashes",
+        "requirements.in",
+        "--allow-unsafe",
+        "--output-file",
+        f"docker/build_scripts/requirements{session.python}.txt",
+    )
+
+@nox.session(python="3.7")
```
mayeut, line 21:
```suggestion
@nox.session(python="3.9")
```
This has been updated between dropping Python 3.5 and adding Python 3.10

henryiii:
No more locale bug for < 3.7 with the manylinux image. :)

henryiii:
Checked it with the update workflow turned on on an earlier build, looks like it's working fine. There might be some changes needed due to the token requirement - see @joerick's workaround on cibuildwheel with a GitHub app token.

mayeut:
> There might be some changes needed due to the token requirement - see @joerick's workaround on cibuildwheel with a GitHub app token.

I've been meaning to ask pypa about a bot / app when saw his work but forgot since then...
That's the last thing to do because otherwise, everything looks good.

henryiii:
pypa/manylinux will need minor changes, I'm looking at https://github.com/pypy/manylinux/blob/8650a6d4a534e72dd3f5d1b38327cc94e404986b/docker/build_scripts_pypy/install_pypy.sh#L46 specifically, which will need the version number added, `requirements.txt` -> `requirements37.txt` and such. Guessing @asottile or @mattip should be aware of the change coming in? I'd be up for making a PR once this goes in if needed.

Wouldn't it be possible to add pypy to the pythons in pypa/manylinux? I think this was hinted at in https://github.com/pypy/manylinux/issues/7, and would solve the issue in https://github.com/pypy/manylinux/issues/15.

mattip:
> Wouldn't it be possible to add pypy to the pythons

Maybe open a separate issue for this?

henryiii:
Actually, it looks like pypy/manylinux is based on an old version of pypa/manylinux; notably it has a copy of requirements.txt internally (which is missing other important things we've added in this repo, including build).

henryiii:
> Actually, it looks like pypy/manylinux is based on an old version of pypa/manylinux; notably it has a copy of requirements.txt internally (which is missing other important things we've added in this repo, including build).

mayeut (approved):
Let's get this in.
I'll deal with the bot a bit later.

PR #1073, opened 2021-04-26
Updated scripts to create manylinux2014 docker image based on centos8…

npanpaliya:
… for ppc64le

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,8 +28,8 @@ jobs:
       env: POLICY="manylinux2014" PLATFORM="aarch64"
     - arch: s390x
       env: POLICY="manylinux2014" PLATFORM="s390x"
-    #- arch: ppc64le
-    #  env: POLICY="manylinux2014" PLATFORM="ppc64le"
+    - arch: ppc64le
+      env: POLICY="manylinux2014" PLATFORM="ppc64le"
     - arch: arm64-graviton2
       virt: vm
       group: edge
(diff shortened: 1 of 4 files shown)


henryiii:
How can centOS 8 make manylinux2014? Wouldn't that be manylinux_2_28?

mayeut:
It would definitely be a manylinux_2_28 image if it ever exists.
CentOS 8 has symbols conflicting with Photon OS 3.0 and thus can't be published as a generic manylinux_2_28 image without further investigations on the likelihood of such conflicts impacting produced wheels (& thus increasing support burden on this project).
Also, this would require updating auditwheel which does not support the manylinux_2_28 policy.

If the manylinux_2_28 policy is added to auditwheel, then, one can try building its wheels on CentOS 8 / Debian 10 like distro.

On a side note, If I were to use CentOS 8, I'd probably use gcc-toolset-10, not gcc-toolset-9.

mayeut:
@npanpaliya,

If you really do want a manylinux_2_28 image based on CentOS 8 then please do open an issue & explain your motivations & rationals for going to this distro.

npanpaliya:
@mayeut - Being a novice user in the docker image world, I've least knowledge about manylinux and its specifications, still learning. I'd thought the names `manylinux_2_28` or `manylinux_2014` are just names, didn't really know what they mean and hence I just added the needed stuff wherever I appropriate for ppc64le. 

My rationale is to create a manylinux image with centos8 as a base image where I can build Tensorflow and family with cuda 11.2

PR #1090, opened 2021-05-10
feat: add symlinks to pythonx.y versions

henryiii:
This is the addition of symlinks mentioned in https://github.com/joerick/cibuildwheel/pull/661; this makes it easier for tools that try to discover multiple Python versions, like Tox/virtualenv/Nox, to use the manylinux images.

A further idea would be to add `python` and `python3` symlinks to a non-system version of Python.


mayeut:
Thanks @henryiii 
There are some unrelated build issue at the moment.
Trying to fix those before merging this.

PR #1095, opened 2021-05-14
Add pipx

mayeut:
Add [pipx](https://pipxproject.github.io/pipx/) to the tools available in manylinux images.
This would help, for example, to run [nox](https://nox.thea.codes/en/stable/) inside the container with a simple command-line now that #1090 is merged in.
Installing pipx rather than nox because it simply allows to run other tools as well.

See also:
#1055 
joerick/cibuildwheel#661


henryiii:
I started to do this, but then worked on #1055 instead since then it's adding one line (though more overall change in generated files).

PR #1103, opened 2021-05-22
Add pypy3.7 7.3.5

### LINKED ISSUE #1099: Add PyPy?
Could the PyPy interpreters (probably starting with the latest one, PyPy3.7 7.3.4 or likely 7.3.5) be added to the manylinux image? This might make it easier to provide different versions (pypy/manylinux is stuck on manylinux 2010 pypy/manylinux#7) and would avoid image issues mentioned in pypy/manylinux#15. 

_see @mattip in https://github.com/pypa/manylinux/issues/1055#issuecomment-842586294_

mayeut:
Fix #1099


docker/build_scripts/finalize.sh
```diff
@@ -28,8 +28,12 @@ for PREFIX in $(find /opt/_internal/ -mindepth 1 -maxdepth 1 -name 'cpython*');
 	ln -s ${PREFIX} /opt/python/${ABI_TAG}
 	# Make versioned python commands available directly in environment.
 	PYVERS=$(${PREFIX}/bin/python -c "import sys; print('.'.join(map(str, sys.version_info[:2])))")
-	ln -s ${PREFIX}/bin/python /usr/local/bin/python${PYVERS}
-	ln -s ${PREFIX}/bin/pip /usr/local/bin/pip${PYVERS}
+	if [[ "${PREFIX}" == *"/pypy"* ]]; then
+		ln -s ${PREFIX}/bin/python /usr/local/bin/pypy${PYVERS}
+	else
+		ln -s ${PREFIX}/bin/python /usr/local/bin/python${PYVERS}
+		ln -s ${PREFIX}/bin/pip /usr/local/bin/pip${PYVERS}
```
mayeut, line 35: I now wonder if the pip link should be removed or not ?

Introduced in #1090 when adding python symlinks.

mayeut:
@mattip, could you review this PR please ?

docker/build_scripts/finalize.sh
mattip: If `pip` is symlinked for cpython, it should be symlinked for pypy too. 

I personally prefer to explicitly use `python -m pip`, but many people seem to prefer to trust that `pip` is properly symlinked.

tests/run_tests.sh
```diff
@@ -24,17 +24,24 @@ for PYTHON in /opt/python/*/bin/python; do
...
+	else
+		LINK_PREFIX=python
+		# Make sure sqlite3 module can be loaded properly and is the manylinux version one
+		# c.f. https://github.com/pypa/manylinux/issues/1030
+		$PYTHON -c 'import sqlite3; print(sqlite3.sqlite_version); assert sqlite3.sqlite_version_info[0:2] >= (3, 34)'
+		# pipx.y shall be available directly in PATH
+		LINK_VERSION=$(pip${PYVERS} -V)
+		REAL_VERSION=$(${PYTHON} -m pip -V)
+		test "${LINK_VERSION%% from *}" = "${REAL_VERSION%% from *}"
+	fi
```
mattip, line 40: Shouldn't these sanity checks be run on PyPy too? If they fail there is a problem with the PyPy build ...

mattip:
Thanks. I plan on releasing PyPy 7.3.5 in the next two days, so maybe merge only after that is released and can be used here?

docker/build_scripts/finalize.sh
mayeut: > If pip is symlinked for cpython, it should be symlinked for pypy too.

But we can't symlink `pip3.7` both for cpython 3.7 and pypy 3.7 in `/usr/local/bin/`

> I personally prefer to explicitly use `python -m pip`, but many people seem to prefer to trust that `pip` is properly symlinked.

I do too.
For the sake of consistency and given the fact that pipx.y were introduced very recently, I'm going to remove them.

tests/run_tests.sh
mayeut: can't run the sanity check of the `pipx.y` symlink in `/usr/local/bin` since there are no symlink for PyPy pip in there. As mentioned in an earlier comment, going to remove those anyway.

For sqlite3, this is specific to cpython being rebuilt from source with a specific version (well should be specific, this has been an issue before). Given PyPY comes as a full featured binary, this check isn't right (& fails because the embedded sqlite3 is version 3.28 or 3.29 - I don't remember exactly)

docker/build_scripts/finalize.sh
mayeut: Thinking back on your comment, I think you might have had `ln -s pip3.7 /opt/python/.../pip` or `ln -s pip3.7 /opt/python/.../pip` in mind rather than adding a link to PyPy's pip in `/usr/local/bin` ?

If that's the case then maybe your comment will change ?

Mine still applies in any case (unless someone objects to removing those pip symlinks in `/usr/local/bin`

tests/run_tests.sh
mattip: > the embedded sqlite3 is version 3.28 or 3.29

Hmm. Is it documented somewhere what sqlite3 a specific python version expects? If cpython is statically linking to 3.35 that would suggest PyPy should as well.

docker/build_scripts/finalize.sh
henryiii: Yes, this was not talking about the pipX.Y, but I'm not attached to those either, pythonX.Y -m pip or pypy3.7 -m pip is better. :)

tests/run_tests.sh
mayeut: The minimum sqlite3 version required by cpython is fairly low: https://github.com/python/cpython/blob/220dd80a2671f57486055955d5422163cf73ed33/Modules/_sqlite/module.c#L33

The default version in manylinux1 was even older than this requirement though:
- https://github.com/pypa/manylinux/issues/90
- https://github.com/pypa/manylinux/pull/93

It's been updated since then in order not to get issue reported because of CVE on older versions.

Windows/macOS installers of CPython also update sqlite3 regularly as noted in their release notes.

docker/build_scripts/finalize.sh
mayeut: `pipX.Y` symlinks in `/usr/local/bin` have been removed in master.
`pythonX.Y -m pip` or `pypyX.Y -m pip` can be used.

mayeut:
> Thanks. I plan on releasing PyPy 7.3.5 in the next two days, so maybe merge only after that is released and can be used here?

Seems like a good plan.
Converted back to draft until updated to 7.3.5 to prevent this PR from being merged.

mattip:
PyPy 3.7.5 has been released. The checksums are available at https://www.pypy.org/checksums.html

mattip:
Is there a way to test that this is working?

mayeut:
> Is there a way to test that this is working?

The tests in the manylinux repo are really minimal & PR images aren't uploaded anywhere.
All I can say is that it passes the same minimal tests as cpython here (mostly, we can install packages with pip and ssl connections are working properly).
I'll merge this once Travis CI is green to get some more tests running in pypa/cibuildwheel#671

YannickJadoul (commented):
Sorry for the delay, @mayeut. I don't know the manylinux images that well, but yeah, this PR seems to make sense to me :-)

PR #1123, opened 2021-06-09
Add pypy3.7 7.3.5 to s390x

messense:
Not sure why #1103 omits s390x, https://downloads.python.org/pypy/ has  PyPy3 s390x release.


henryiii:
@mayeut is out for a couple of weeks, I think he's back next week. Looks like from the Travis run that `libncursesw.so` is missing for s390x.

mayeut:
@messense, the relevant discussion for #1103 not including s390x is there: https://github.com/pypa/manylinux/issues/1099#issuecomment-843429187 

In short, the s390x binaries of PyPy are not portable across distros.

I honestly thought it would work on manylinux2014 but the failure in CI tends to prove otherwise.

PR #1159, opened 2021-08-03
chore: bump to Python 3.10.0rc1

henryiii:


henryiii:
Interesting, Python 3.10rc1 seems to be broken on manylinux2010 where beta4 worked.  (Travis failure looks like a normal flake).

mayeut:
As mentioned in the fix commit, I have no idea why it started failing on manylinux2010.
I do not have much time right now to report this or do further analysis & I'll be away for 2 weeks so I'd rather have a quick workaround for this than not having rc1 for now.

PR #1161, opened 2021-08-06
Update patchelf to 0.13

lpsinger:


auvipy (approved):
is the travis failures related?

lpsinger:
I don't think so.

lpsinger:
When will this be merged? One of our packages is affected by a bug in patchelf that was fixed in version 0.13 (see https://git.ligo.org/lscsoft/lalsuite/-/issues/459#note_409265) and we cannot release until this is fixed in manylinux.

lpsinger:
Thanks!

PR #1167, opened 2021-08-21
chore: add native dependency update script

mayeut:
Add native dependency updates in the update dependencies workflow.
Each native tool updated has its own commit.


marcelotduarte:
The "Update the requirements" session of [readme](https://github.com/pypa/manylinux#updating-the-requirements) needs to be updated too.

Issue #1149, opened 2021-07-28
`--enable-shared` is required by Pyinstaller


ikus060:
According to PyInstaller https://github.com/pyinstaller/pyinstaller/issues/6066#issuecomment-888333161 manylinux images could not be used to create redistributable package compatible with many linux distribution because the python provided within the images are not compiled with `--enable-shared`

Mostly a question or a feature request, what would be the properly way to have a "manylinux" image with  `--enable-shared` ?

mayeut:
This was done intentionnaly for building wheels which is the use-case of manylinux images.
c.f. https://github.com/pypa/manylinux/issues/255#issuecomment-450761634 & #30 for more details.

If I'm not mistaken, the use-case for pyinstaller only requires 1 python installation to create the "app" package.
This can be done in a "manylinux" friendly way by installing python3-devel on a CentOS base image.
CentOS 7 (base image for manylinux2014) provides 2 versions of python3:
- 3.6.8 through `yum install python3-devel`
- 3.8.6 through `yum install rh-python38-python-devel` (you need to enable `scl-rh` repo for this one: `yum -y install centos-release-scl-rh`)

IMHO, this is the best option for pyinstaller. It also allows not to waste bandwidth/ci time using images that are larger than necessary.

ikus060:
Thanks alot for the information.

Will use manylinux2014 base image and install the required devel package to make it work.

Issue #793, opened 2020-10-05
Linking to libpython


alex-eri:
I want to build cx-Freeze against manylinux, but
```
    gcc -pthread -Wno-unused-result -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -fPIC -I/opt/_internal/cpython-3.7.9/include/python3.7m -c source/bases/Console.c -o build/temp.linux-x86_64-3.7/source/bases/Console.o
    creating build/lib.linux-x86_64-3.7/cx_Freeze/bases
    gcc -pthread build/temp.linux-x86_64-3.7/source/bases/Console.o -L/opt/_internal/cpython-3.7.9/lib/python3.7/config-3.7m-x86_64-linux-gnu -lpython3.7m -o build/lib.linux-x86_64-3.7/cx_Freeze/bases/Console -Xlinker -export-dynamic -lcrypt -lpthread -ldl -lutil -lrt -lm -s
    /opt/rh/devtoolset-8/root/usr/libexec/gcc/x86_64-redhat-linux/8/ld: cannot find -lpython3.7m
    collect2: error: ld returned 1 exit status
    error: command 'gcc' failed with exit status 1

```

madig:
I think you're not supposed to link to libpython: https://www.python.org/dev/peps/pep-0513/#libpythonx-y-so-1.

mayeut:
duplicate of https://github.com/pypa/manylinux/issues/1149

Issue #255, opened 2019-01-01
libpython*.so is missing


dmitriy-serdyuk:
Python installation seems to be strange, there is no `libpython.so`:

```
$ docker exec cont  /opt/python/cp37-cp37m/bin/python -c "from distutils import sysconfig; print(sysconfig.get_config_var('LIBDIR'))"
/opt/_internal/cpython-3.7.1/lib
$ docker exec 5319c1eb2c41 ls /opt/_internal/cpython-3.7.1/lib
pkgconfig
python3.7
$ docker exec 5319c1eb2c41  find / -name 'libpython3.7*'
```

Am I doing something wrong?

rdb:
There isn't supposed to be.  You do not need to link extension modules against libpython.so.

itamarst:
I have the rather obscure use case of wanting to build Python binaries that link against libpython using the manylinux image. So this lack hit me too, but also my use case may be too obscure to be worth your time fixing.

rdb:
@itamarst See #91 for discussion about that use case.

itamarst:
Ah, interesting.

In my case however I'd like to actually link against libpython.so, rather than .a, so the Python used matches the patch version on the user's computer. I'm basically building a lightweight Python binary as an alternative to LD_PRELOADing; I'm assuming the user already has a Python install on their computer. Again, somewhat obscure use case :)

rdb:
I'd say that building things that are decidedly not manylinux-compatible fall outside the scope of the manylinux project.  You can make your own Docker image based on manylinux that builds a new copy of Python with shared libraries enabled.  You will have to do that anyway, as manylinux builds Python statically, so no libpython.so is created at any point during the process.

Keep in mind that "having a Python install on your computer" is no guarantee of having libpython.so available.  Your produced binaries, despite not qualifying for the manylinux tags, will not work with any static distribution of Python.

dmitriy-serdyuk:
@itamarst you can try to build a new image commenting out this line https://github.com/pypa/manylinux/blob/master/docker/build_scripts/build.sh#L187

Issue #30, opened 2016-02-06
Missing libpython with Debian default Python install


matthew-brett:
I am testing manylinux numpy wheels.

In particular, I am testing this guy: http://nipy.bic.berkeley.edu/manylinux/numpy-1.10.4-cp27-none-linux_x86_64.whl

With a default install, test of Python, starting with either Wheezy or Jessie:

```
docker run -ti --rm -v $PWD:/io debian:latest /bin/bash
docker run -ti --rm -v $PWD:/io tianon/debian:wheezy /bin/bash
```

and running this script:

```
apt-get update 
apt-get install -y python curl 
curl -sLO https://bootstrap.pypa.io/get-pip.py 
python get-pip.py 
pip install -f https://nipy.bic.berkeley.edu/manylinux numpy nose 
python -c "import numpy; numpy.test()" 
```

I get this:

```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/usr/local/lib/python2.7/dist-packages/numpy/__init__.py", line 180, in <module>
    from . import add_newdocs
  File "/usr/local/lib/python2.7/dist-packages/numpy/add_newdocs.py", line 13, in <module>
    from numpy.lib import add_newdoc
  File "/usr/local/lib/python2.7/dist-packages/numpy/lib/__init__.py", line 8, in <module>
    from .type_check import *
  File "/usr/local/lib/python2.7/dist-packages/numpy/lib/type_check.py", line 11, in <module>
    import numpy.core.numeric as _nx
  File "/usr/local/lib/python2.7/dist-packages/numpy/core/__init__.py", line 14, in <module>
    from . import multiarray
ImportError: libpython2.7.so.1.0: cannot open shared object file: No such file or directory
```

Sure enough, in order for the wheel to work, I need:

```
apt-get install libpython2.7
```

Maybe we need to add a check / informative error message for the presence of libpython?

matthew-brett:
Same on Ubuntus 14.04, 15.10.

rmcgibbo:
Where would such a check go? This is really a Debian problem with the way they split upstream python into many subpackages,  no?

matthew-brett:
It would have to go in the pip check for the validity of manylinux for the platform.

It is unfortunate that Debian does this, but I bet that a large majority of our consumers will be on Debian or Ubuntu, and there's not much prospect of Debian / Ubuntu changing this immediately, and even if they do, there will be many systems liable to this problem in the wild.

[...]

PR #1185, opened 2021-09-15
docker: Add ability to build shared library flavours of CPython

rcoup:
Add a build option to enable building shared/libpython-enabled Python. Helps a number of use-cases (eg: #1149 #793 #255 #30). In mine, using PyInstaller for creating a works-on-many-linux distribution.

Obviously as described in some of the above tickets, it doesn't fit the manylinux requirements and shouldn't be on by default. But adding the option helps people reuse the huge work put into manylinux images without maintaining forks or redoing it all.

Enable via:

    MANYLINUX_BUILD_EXTRA="--build-arg PY_SHARED=1" \
    PLATFORM=$(uname -m) POLICY=manylinux2014 COMMIT_SHA=latest \
    ./build.sh

If set, builds both static & shared versions of CPython interpreters into the images.

Shared versions end up in (eg) `/opt/python/cp37-cp37m-shared/` alongside the existing static `/opt/python/cp37-cp37m/`, with a Python binary as `/usr/local/bin/python3.7-shared`.


rcoup:
If people are happy enough I will update the README/etc too. And could even expose it as a slightly less hidden top-level `PY_SHARED=1` option.

auvipy (approved):
looks good to me

rcoup:
> If people are happy enough I will update the README/etc too. And could even expose it as a slightly less hidden top-level PY_SHARED=1 option.

I've done that in the latest push.

docker/build_scripts/finalize.sh
```diff
@@ -24,14 +30,16 @@ for PREFIX in $(find /opt/_internal/ -mindepth 1 -maxdepth 1 \( -name 'cpython*'
 	# Since we fall back on a canned copy of pip, we might not have
 	# the latest pip and friends. Upgrade them to make sure.
 	${PREFIX}/bin/pip install -U --require-hashes -r ${MY_DIR}/requirements${PY_VER}.txt
+
 	# Create a symlink to PREFIX using the ABI_TAG in /opt/python/
+
 	ABI_TAG=$(${PREFIX}/bin/python ${MY_DIR}/python-tag-abi-tag.py)
-	ln -s ${PREFIX} /opt/python/${ABI_TAG}
+	ln -s ${PREFIX} /opt/python/${ABI_TAG}${SUFFIX:-}
```
mayeut, line 37:
```suggestion
	ln -s ${PREFIX} /opt/python/${ABI_TAG}${SUFFIX}
```

mayeut (changes requested):
Thanks for taking the time to do this.
Before this can be merged, please do explain why you'd need a manylinux like image. The issues you mentioned provides another way to handle the PyInstaller/cx-freeze use-case with minimal effort. What use-case is not solved with a minimal "CentOS" (or other base image) image with python ?
I'd rather understand the added value for downstream repos before including bits of code that will never be tested here.

termim:
>  Before this can be merged, please do explain why you'd need a manylinux like image. 

To build and test PyInstaller/cx-freeze based executables for every supported python version.
 
> The issues you mentioned provides another way to handle the PyInstaller/cx-freeze use-case with minimal effort. 

No, the issues mentioned basically say - use whatever python provided by the base image or build your own. And manylinux project itself proves that it is not a minimal effort.

> What use-case is not solved with a minimal "CentOS" (or other base image) image with python ? 

The use case is pretty obvious - [base your work on top of well done manylinux project](https://en.wikipedia.org/wiki/Standing_on_the_shoulders_of_giants) and avoid reinventing this wheel (properly build all supported python versions) over and over again.

> I'd rather understand the added value for downstream repos before including bits of code that will never be tested here.

The added value for downstream repos is ability to use properly built all supported python versions with up to date supporting libraries like openssl etc. for building/testing executables from python scripts.

mayeut:
> To build and test PyInstaller/cx-freeze based executables for every supported python version.

@termim, can you please explain the use-case (or better, provide a real reference for the use-case) of building an application/executable with every supported python version ? This is where I'm lost. If you're building an application, only one python version should be enough ?

Not sure if it's been mentioned before but there's also https://github.com/indygreg/python-build-standalone/ python builds which are meant to be highly portable builds of python (manylinux just build without any care that the python builds themselves should be portable, only the built wheels are taken care of).

black-sliver:
:+1: for /opt/python/cp39-cp39-shared.
My use-case would be that the project I want to bundle in an AppImage uses (and kind of depends on) cx_Freeze and is going to move to py3.9+ soon. I chose the AppImage path specifically because it would allow users to run it on older OS with older glibc that do not ship the latest and greatest python.

AlmogBaku:
Any update on this?

pkit:
Python community is notoriously sloppy with how glibc dependencies are handled.
This project is the only place where some sanity is preserved.
Not having `libpython.so` though is a major dealbreaker for people who need a stable python environment, with clearly defined glibc dependency.
"Do it yourself" is not a good idea, because in the end it will be yet another manylinux just less used/tested.
You can look into [`aws-cli`](https://github.com/aws/aws-cli/tree/v2/scripts) for amount of hoops people need to jump.

aurelien35:
I ran into the same problem: the need for a manylinux with shared libraries available.

I understand that build shared libraries cannot be the default behaviour, but I think an option to support it as in this pull request is a good idea.


Is there any plan to accept this pull request?

PR #1198, opened 2021-09-27
Mention 3.10 in docs

tonybaloney:
manylinux2014 does have 3.10 inside, so update the README to reflect this


auvipy:
3.10 is not general available yet

tonybaloney:
> 3.10 is not general available yet

sure, but it is in the image and folks are wanting to test it before it goes GA (myself included)

mayeut:
Thanks for the update to the docs.
I removed a mention of 3.10.0+ that was redundant with 3.9.0+

PR #1211, opened 2021-10-21
ci: emulate ppc64le on GHA

henryiii:


henryiii:
That is really weird - the exact same build passes in my fork, https://github.com/henryiii/manylinux/actions/runs/1369744930, and the error in the one failure is one that the second commit explicitly fixed.

henryiii:
I think I found the problem, the shell is sh, not bash.

auvipy (approved):

henryiii:
@auvipy or @mayeut (if this is still a problem when you return early November): the upload authentication seems to be scoped to the specific images, so the three moved images are not being uploaded because the credentials are not registered. GHA should be added to these images. I can also make a revert PR so we can monitor to see when Travis starts working again (haven't heard back from them).

PR #1222, opened 2021-11-07
ci: move ppc64le builds back to travis-ci

mayeut:
ppc64le travis-ci seems to be working with the docker-buildx frontend now. on both toronto & saopolo ppc64le workers.
Let's get those builds there (for now)  to remove slow qemu builds from GHA.
Keeping (for now) the (currently) unneeded bits for other frontends to ease future debugs. They've all been tested and work to some extent (depending on worker - saopolo or toronto, the docker one has no cache). I could add a commit to drop them and still be able to find them but don't want to skim through all commits if this is ever needed.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -33,7 +33,7 @@ jobs:
       fail-fast: false
       matrix:
         policy: ["manylinux2014", "manylinux_2_24", "musllinux_1_1"]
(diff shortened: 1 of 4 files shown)


henryiii:
Is this related to https://github.com/conda-forge/conda-smithy/pull/1520 perhaps?

PR #1236, opened 2021-12-12
chore: use PyPI patchelf rather than rebuilding from source

mayeut:
patchelf 0.14+ requires c++17
c++17 is not available on all manylinux images, use a prebuilt patchelf wheel for installation.

Before merging this in, I'd like to have the input of some PyPA members on the added dependency and, in order not to be the only one with access to the patchelf package on PyPI, add them as maintainers to both PyPI & GitHub projects:
- https://pypi.org/project/patchelf/
- https://github.com/mayeut/patchelf-pypi


mayeut:
@henryiii, @lkollar, any thoughts on this one ?

lkollar:
I think this is definitely an improvement over the current situation. I'm happy to help out.

PR #1239, opened 2021-12-14
Add support for Pyston 2.3.1

undingen:
This commit integrates [Pyston](https://github.com/pyston/pyston) into manylinux.
While Pyston 2.x is based on CPython 3.8 it's not ABI compatible so recompiling of binary extensions is required. Providing precompiled pip installable packages would greatly help our users.

- Pyston currently only runs on `x86_64`
- gets installed into `/opt/python/pyston38-pyston_23_x86_64_linux_gnu` ->
  `/opt/_internal/pyston3.8-2.3.1`
- `manylinux2014` & `manylinux_2_24` supported
- EOL `manylinux2010` isn't supported because the supplied glibc is too old for Pyston
- `musllinux` is not supported

With this changes and the my WIP changes to [cibuildwheel](https://github.com/undingen/cibuildwheel/tree/pyston_2.3.1) I was able to generate wheels for Pyston.

Notes:
- Please very carefully review my PR because this is the first time I'm working with manylinux.
- Patch is mainly copied from the script for PyPy.
- We don't have a [PEP425](https://www.python.org/dev/peps/pep-0425/) Python Tag yet.
- Compiling Pyston is not really feasible because it takes many hours and is only supported on Ubuntu and has quite a few additional dependencies that's why I'm using the binary package.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -147,7 +147,7 @@ Image content
 
 All images currently contain:
 
-- CPython 3.6, 3.7, 3.8, 3.9, 3.10, and PyPy 3.7, 3.8 installed in
+- CPython 3.6, 3.7, 3.8, 3.9, 3.10, and PyPy 3.7, 3.8 and Pyston 2.3 installed in
   ``/opt/python/<python tag>-<abi tag>``. The directories are named
   after the PEP 425 tags for each environment --
   e.g. ``/opt/python/cp37-cp37m`` contains a CPython 3.7 build, and
@@ -167,6 +167,8 @@ became useless (builds with and without pymalloc are ABI compatible) and so has
 been removed. (e.g. ``/opt/python/cp38-cp38``)
 
 Note that PyPy is not available on ppc64le & s390x.
+
+Note that Pyston is only available on x86_64.
 
 Building Docker images
 ----------------------
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -143,8 +143,11 @@ RUN manylinux-entrypoint /build_scripts/build-cpython.sh 3.10.0
 FROM build_cpython AS all_python
 COPY build_scripts/install-pypy.sh /build_scripts/install-pypy.sh
 COPY build_scripts/pypy.sha256 /build_scripts/pypy.sha256
+COPY build_scripts/install-pyston.sh /build_scripts/install-pyston.sh
+COPY build_scripts/pyston.sha256 /build_scripts/pyston.sha256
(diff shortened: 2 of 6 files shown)


undingen:
I have started modifying `wheel` to return `pt38-pt23` for `print("{0}-{1}".format(tag.interpreter, tag.abi))` in order
to change the install location to `/opt/python/pt38-pt23`.
https://github.com/undingen/wheel/tree/pyston_2.3.1

I guess it makes sense to first try to get the shortcut `pt` for `pyston` accepted and only than add Pyston to pypa. 
What do you folks think?

auvipy (approved):
i am ok with it. lets see what other maintainers suggest

mayeut (changes requested):
@undingen,

The PR itself does look okay except for the 2 comments I left.
What worries me a bit more is the following:
> We don't have a PEP425 Python Tag yet.

> I have started modifying wheel to return pt38-pt23 for print("{0}-{1}".format(tag.interpreter, tag.abi)) in order
to change the install location to /opt/python/pt38-pt23.
https://github.com/undingen/wheel/tree/pyston_2.3.1
I guess it makes sense to first try to get the shortcut pt for pyston accepted and only than add Pyston to pypa.
What do you folks think?

There are a few things to take into account when building/publishing/consuming manylinux wheels.
Let's have a look at the consuming part to see how changing python/ABI tags will impact this:
I will take a shortcut here and only consider `pip install package`
For now, `pyston3 -m pip debug` returns 215 compatible tags like:
```
Compatible tags: 215
  pyston38-pyston_23_x86_64_linux_gnu-manylinux_2_17_x86_64
  pyston38-pyston_23_x86_64_linux_gnu-manylinux2014_x86_64
  pyston38-pyston_23_x86_64_linux_gnu-manylinux_2_16_x86_64
...
```
If anything changes in the python/ABI tags, this will require an update to `pip` (through [packaging](https://github.com/pypa/packaging)) and all users to update `pip` in order to be able to install PyPI published packages.
This also means that Pyston should release new binaries with an updated `pip` if they want their users to be able to install wheels out-of-the-box.
`packaging` is one of the packages you patched in your `wheel` fork (thus `wheel` will not accept the patch as-is).
You'd probably want backward compatibility so make the patch to `packaging` generate the older python/ABI tags as well.

I encourage you to start a discussion on https://discuss.python.org/c/packaging/14 as most choices will impact many packages (might be better to get acceptability of such choices) and maybe Pyston itself.

IMHO, this shall not be merged until a decision about ABI tags has been made.

mayeut:
too old, closing.

PR #1265, opened 2022-01-25
Add ability to apply patches for CPython

### LINKED ISSUE #1264: Python.h:44:10: fatal error: crypt.h: No such file or directory
Image: manylinux_2_24
Additional installed packages: libxml2
Error:
```
/opt/python/cp37-cp37m/include/python3.7m/Python.h:44:19: fatal error: crypt.h: No such file or directory
     #include <crypt.h>
                       ^
    compilation terminated.
```
About the issue: https://bugs.gentoo.org/802210
Resolved Python issue: https://bugs.python.org/issue44751

Question: Does Python in manylinux images include this fix?

MarshalX:
close #1264

Key notes:
- patches was generated via `diff -Naur` with hand edits (for example I deleted timestamps) 
- patches applied only for `manylinux_2_24` images
- patches stored in separated folder
- patches applied silently
- to control patch applying `maybe-patch-cpython.sh` script was created. It allows control which patches should be applied by CPython version 
- there is only `bpo-44751` patch for cp36, cp37 and cp38 for now

Also I deploy patched images for easy tests (and my usage in pet-project) here: https://quay.io/organization/pypa_patched

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -107,7 +107,11 @@ RUN export OPENSSL_ROOT=openssl-1.1.1m && \
     manylinux-entrypoint /build_scripts/build-openssl.sh
 
 COPY build_scripts/build-cpython.sh /build_scripts/
+COPY build_scripts/maybe-patch-cpython.sh /build_scripts/
+COPY build_scripts/cpython-patches /build_scripts/cpython-patches
 
+# We should grant exec perrmission because it will be called not from Dockerfile
+RUN chmod +x /build_scripts/maybe-patch-cpython.sh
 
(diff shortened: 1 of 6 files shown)


mayeut:
I really do not wish for us to patch CPython as this adds maintenance burden. If the bug was serious enough, I guess it would have been fixed upstream earlier.
The PR itself raises many questions, like why manylinux_2_24 only ? what about PyPy ?
As such I will close this PR.

Thanks for putting an effort into it @MarshalX, but this won't go in. I'd rather spend a little bit more time discussing why the specific issue fixed here is a real one than get this maintenance burden in for "the rest of time".

PR #1278, opened 2022-02-06
Use vault for repos in manylinux2014

alex:
Starting a few days ago builds that use yum began failing like so: https://github.com/pyca/infra/runs/5074957413?check_suite_focus=true

### DIFF
--- a/docker/build_scripts/fixup-mirrors.sh
+++ b/docker/build_scripts/fixup-mirrors.sh
@@ -4,7 +4,7 @@
 # Stop at any error, show all commands
 set -exuo pipefail
 
-if [ "${AUDITWHEEL_POLICY}" == "manylinux2010" ]; then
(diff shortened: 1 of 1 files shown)


alex:
Apparently this is not correct. I'd assumed that the fix here was the same as for manylinux2010, but apparently not.

Does anyone recognize teh failure mode and know how to fix?

mayeut:
@alex, 
CentOS 7 is not EOL yet so there's no reason to move to the vault yet. 
The log linked mentions the EPEL repo is the culprit, I'll try to have a look (I've not experienced any failures for manylinux2014 yet when building manylinux images).

alex:
Hmm, interesting. I guess I got confused due to the CentOS8 derepcation (in favor of Stream). I clearly do not understand the CentOS release system :-)

Thanks for taking a look.

mayeut:
could not find anything relevant, left a comment for possible mitigation in https://github.com/pyca/infra/pull/384

PR #1289, opened 2022-02-24
do not use pypy3.7-v7.3.8

mattip:
PyPy3.7 broke binary compatibility with the latest release. Pin to the older version. Also add checksums for aarch64. See https://foss.heptapod.net/pypy/pypy/-/issues/3688 for more details


auvipy (approved):

mayeut:
Thanks for the PR @mattip 
I hacked the `portable` update in such a way that it broke when reverting to 7.3.7
Not sure what the future is made of so for now, the `-portable` suffix is only added for aarch64 7.3.8
It might take a while for CI to give us the green light but I hope it should be good now.

mattip:
Thanks for the fix, sorry for the mess

Issue #1317, opened 2022-05-14
Add CPython 3.11


mayeut:
With CPython 3.11.0 beta1 available, it's time to add CPython 3.11 to manylinux images.

The issue will be closed once CPython 3.11.0 is released.

EDIT: 2022-08-21

CPython 3.11.0 rc1 is now available in all `manylinux2014`, `manylinux_2_24`, `manylinux_2_28` & `musllinux_1_1` images.

mayeut:
CPython 3.11.0 beta1 is now available in all `manylinux2014`, `manylinux_2_24` & `musllinux_1_1` images.

The issue will be closed once CPython 3.11.0 is released.

dhdaines:
I think this change is breaking my builds, see #1347

dhdaines:
Is it possible to revert to the previous version of the image somehow?  Sorry, I am not very experienced with Docker...

mayeut:
3.11.0 addede in #1399

PR #1318, opened 2022-05-14
feature: add CPython 3.11

mayeut:
relates to #1317


mayeut:
Failure on manylinux 2010 is related to https://github.com/python/cpython/issues/91731#issuecomment-1121173917

PR #1345, opened 2022-07-10
Drop support for manylinux2010

### LINKED ISSUE #1281: Drop support for manylinux2010 images August 1st, 2022
CentOS 6 support - which is used as the base image for manylinux2010 - has been EOL for more than a year.

We're seeing regular issues during image builds with yum accessing the vault repositories and this might become "unsustainable" (for manylinux maintainers and for packagers with a need to install system packages with `yum`).

There's now very little need for `manylinux2010` compatible wheels where the "pip readiness" is less of an issue than with the drop of `manylinux1`: the highest percentage is for python 3.7 with around 7% of systems that are >= manylinux2010 and < manylinux2014, due to the pip version installed only.

This will happen this year, how soon will probably depend on build stability.
Images won't be removed from quay.io and will still be usable as-is.

discourse thread: https://discuss.python.org/t/dropping-manylinux2010-docker-images-support-in-2022/13540

mayeut:
closes #1281

### DIFF
--- a/docker/build_scripts/fixup-mirrors.sh
+++ b/docker/build_scripts/fixup-mirrors.sh
@@ -3,12 +3,3 @@
 
 # Stop at any error, show all commands
 set -exuo pipefail
-
-if [ "${AUDITWHEEL_POLICY}" == "manylinux2010" ]; then
-	# Centos 6 is EOL and is no longer available from the usual mirrors, so switch
-	# to https://vault.centos.org
-	sed -i 's/enabled=1/enabled=0/g' /etc/yum/pluginconf.d/fastestmirror.conf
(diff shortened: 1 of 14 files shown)


docker/build_scripts/fixup-mirrors.sh
```diff
@@ -3,12 +3,3 @@
 
 # Stop at any error, show all commands
 set -exuo pipefail
```
h-vetinari, line 5: Sounds like the rest of that file could be deleted too.

mayeut: Thanks for the suggestion.

The whole fixup-mirrors could be removed however:
- This line is not manylinux2010 specific so not removed. It's part of the skeleton of every script in here.
- I'd rather keep the file as it will be needed at some point (I'd rather have it kept around rather than dig through history).

Issue #1306, opened 2022-04-04
Publish multi-platform images


reuben:
It'd be great if the project could publish multi-platform images so that one can do multi-arch Docker builds using buildx and e.g. `docker buildx build --platforms linux/amd64,linux/arm64`.

See some relevant docs:
 - https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/
 - https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/

Right now, because the images for different architectures have different names, one either has to create a manifest by hand, or use separate downstream Dockerfiles for each arch. I've tried out with my own manifest and it works well for multi-platform builds:

```
$ docker pull quay.io/pypa/manylinux_2_24_aarch64:2022-03-31-361e6b6
$ docker pull quay.io/pypa/manylinux_2_24_x86_64:2022-03-31-361e6b6
$ docker tag quay.io/pypa/manylinux_2_24_x86_64:2022-03-31-361e6b6 ghcr.io/reuben/manylinux_2_24_x86_64:2022-03-31-361e6b6
$ docker tag quay.io/pypa/manylinux_2_24_aarch64:2022-03-31-361e6b6 ghcr.io/reuben/manylinux_2_24_aarch64:2022-03-31-361e6b6
$ docker login ghcr.io
$ docker push ghcr.io/reuben/manylinux_2_24_x86_64:2022-03-31-361e6b6
$ docker push ghcr.io/reuben/manylinux_2_24_aarch64:2022-03-31-361e6b6
$ docker manifest create ghcr.io/reuben/manylinux_2_24:2022-03-31-361e6b6 \
  --amend ghcr.io/reuben/manylinux_2_24_x86_64:2022-03-31-361e6b6 \
  --amend ghcr.io/reuben/manylinux_2_24_aarch64:2022-03-31-361e6b6
$ docker manifest push ghcr.io/reuben/manylinux_2_24:2022-03-31-361e6b6
```

The above lets me write a Dockerfile with `FROM ghcr.io/reuben/manylinux_2_24:2022-03-31-361e6b6` and build it for multiple platforms with a single `docker buildx build --platforms linux/amd64,linux/arm64 .` call.

mayeut:
Thanks @reuben for this proposal.
I'm open to review a PR implementing this.
[...]

PR #1412, opened 2022-11-07
Deploy multi-platform images

messense:
#1306


deploy_multi_platform.sh
```diff
@@ -0,0 +1 @@
+#!/bin/bash
```
messense, line 1: TODO: write the script to actually create the multi-platform docker manifest.

.github/workflows/build.yml
```diff
@@ -73,3 +73,27 @@ jobs:
...
+    steps:
+      - name: Wait for Travis CI
+        uses: fountainhead/action-wait-for-check@v1.1.0
+        id: wait-for-travis
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          checkName: Travis CI - Pull Request
+          ref: ${{ github.event.pull_request.head.sha || github.sha }}
+          intervalSeconds: 10
+          timeoutSeconds: 3600
```
mayeut, line 91: Unfortunately, I expect the Travis CI check to fail every now and then (appart from the timeout being way too short for a Travis CI build).

This should really be a different workflow, possibly triggered by the checks being updated on main ?

messense: A different workflow should be fine, but it's harder to test since it can only run after merging. 

You could extract it to a separate workflow afterwards.

mayeut: It's fine to leave this here at the moment as you're saying.
It shouldn't be to hard to split with the deployment script being a separate bash script.

PR #1415, opened 2022-11-13
feat: use gcc-toolset-12 on manylinux_2_28

mayeut:
use gcc-toolset-12 on manylinux_2_28

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -28,7 +28,7 @@ concurrency:
 jobs:
   build_manylinux:
     name: ${{ matrix.policy }}_${{ matrix.platform }}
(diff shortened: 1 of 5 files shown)


mayeut:
`/opt/_internal/cpython-3.6.15/bin/python -m ensurepip` segfaults on x86_64, more investigation needed on this.

mayeut:
given "-O3" was already used for building wheels with gcc-11, the fix needed here for python 3.6 shall hopefully have minimal impact on wheel building. It will be possible to change the default `OPT` flags on python 3.6 if there are reports of things going bad with this new compiler.

auvipy (approved):

PR #1518, opened 2023-08-13
Rework PyPy installation

mayeut:
PyPy installation is just a tar extract.

As such, it might be interesting to install PyPy at runtime rather than at build time.
This is especially true for EOL versions of PyPy which would allow to reduce the image size, and thus, overall build time for users not using EOL PyPy versions.

This PR does not modify default installed PyPy versions yet.

With this revisited installation method, it will be much easier to add GraalPy.

The argument to `manylinux-interpreters ensure` shall be the most specific PEP425 `{python_tag}-{abi_tag}` for a given interpreter to ease cibuildwheel integration in a first step (given cibuildwheel uses `/opt/python/{python_tag}-{abi_tag}`, should be easy to add a call to `manylinux-interpreters ensure` there).


mayeut:
@pypa/cibuildwheel-maintainers,

Any thoughts on this `manylinux-python ensure`  way of making sure a given interpreter is made available ?

xref https://github.com/pypa/cibuildwheel/pull/1538

joerick (commented):
I think the approach generally seems good. Probably most packagers are getting manylinux only for CPython, so deferring the other interpreters to runtime makes sense, and will cut down on size and download time.

Any numbers on how long the install of an interpreter takes? The interpreters themselves look like just tarfile extracts, but there's also the pip install bit.

We'd want to get manylinux-python support into cibuildwheel before the rollout of this feature I suppose. Within cibuildwheel, if pypy was excluded, I can see a potential issue in our integration test suite, because a few tests do use pypy. The pypy install would happen on each test run, rather than being reused from previous runs. This would increase the CI time quite a bit, I think (and be wasteful on bandwidth).

(Just thinking.... another idea could be a manylinux-slim series of images, which contain cpython only, and a set of manylinux-full images, which use manylinux-slim as a BASE and install all the interpreters on top. No idea how much work it would be to make that happen, but would be quite neat. cibuildwheel could choose the slim variants if it only needs cpython, but use the full ones in the tests)

docker/build_scripts/python_versions.json
```diff
@@ -0,0 +1,82 @@
...
+      "version": "23.0.0",
+      "download_url": "https://github.com/oracle/graalpython/releases/download/graal-23.0.0/graalpython-23.0.0-linux-amd64.tar.gz",
+      "sha256": "25e4fa7c1d45db6dcac5bfa4d1a0aa9ef5581623dc5903ce98d246c5d394639c"
+    },
+    "aarch64": {
+      "version": "23.0.0",
+      "download_url": "https://github.com/oracle/graalpython/releases/download/graal-23.0.0/graalpython-23.0.0-linux-aarch64.tar.gz",
+      "sha256": "e2a00b2b6485282b4a04aa382e30d696e00d20eb2fe1736debbe2d9df2a8737a"
+    }
+  },
```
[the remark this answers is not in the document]
mayeut: > Is the intention that manylinux would still 'pin' specific versions of graalpy 3.10? That seems to be how it looks here.

yes, still pin versions of graalpy/pypy.

docker/build_scripts/manylinux-python.py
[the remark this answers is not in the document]
mayeut: > since this is a user-facing script, might be worth workshopping the name a bit?

Indeed. Thanks for the suggestions.
`manylinux-interpreters` seems like a better fit I think.

> Also, I assume this script would be used by musllinux too?

yes, though mostly useless at the moment, it would be available with the same name (as you said, the repo name rather than the image name).

docker/build_scripts/python_versions.json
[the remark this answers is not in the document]
mayeut: I'd rather have this file be an implementation detail and have manylinux-inspector rely on `manylinux-python list -v`.
It was mostly made to test things for now with something more human readable than machine readable but it could be enhanced to output a json file rather than this:
```
[root@28c69824165a /]# manylinux-python list -v
cp36-cp36m:
  installed:            yes
  python version:       3.6.15
  implemention:         cpython
  implemention version: 3.6.15
cp37-cp37m:
  installed:            yes
  python version:       3.7.17
  implemention:         cpython
  implemention version: 3.7.17
cp38-cp38:
  installed:            yes
  python version:       3.8.17
  implemention:         cpython
  implemention version: 3.8.17
cp39-cp39:
  installed:            yes
  python version:       3.9.17
  implemention:         cpython
  implemention version: 3.9.17
cp310-cp310:
  installed:            yes
  python version:       3.10.12
  implemention:         cpython
  implemention version: 3.10.12
cp311-cp311:
  installed:            yes
  python version:       3.11.4
  implemention:         cpython
  implemention version: 3.11.4
cp312-cp312:
  installed:            yes
  python version:       3.12.0rc1
  implemention:         cpython
  implemention version: 3.12.0rc1
graalpy310-graalpy230_310_native:
  installed:            no
  python version:       3.10
  implemention:         graalpy
  implemention version: 23.0.0
pp37-pypy37_pp73:
  installed:            yes
  python version:       3.7.13
  implemention:         pypy
  implemention version: 7.3.9
pp38-pypy38_pp73:
  installed:            yes
  python version:       3.8.16
  implemention:         pypy
  implemention version: 7.3.11
pp39-pypy39_pp73:
  installed:            yes
  python version:       3.9.17
  implemention:         pypy
  implemention version: 7.3.12
pp310-pypy310_pp73:
  installed:            yes
  python version:       3.10.12
  implemention:         pypy
  implemention version: 7.3.12
```

mayeut:
> Any numbers on how long the install of an interpreter takes? The interpreters themselves look like just tarfile extracts, but there's also the pip install bit.

I don't have them right now but you're right, there's also the pip install bit, it's worth figuring this out now. I'll try this out on GHA (my connection is so slow that it wouldn't be representative of CI usage).

> We'd want to get manylinux-python support into cibuildwheel before the rollout of this feature I suppose. Within cibuildwheel, if pypy was excluded, I can see a potential issue in our integration test suite, because a few tests do use pypy. The pypy install would happen on each test run, rather than being reused from previous runs. This would increase the CI time quite a bit, I think (and be wasteful on bandwidth).

Indeed, we'd want to get manylinux-python support into cibuildwheel before removing any pre-installed PyPy versions.
This order would thus be:
1. merge this PR to roll-out manylinux-python 
2. get manylinux-python support into cibuildwheel
3. remove EOL PyPy versions from the default image

As for cibuildwheel tests, this can be worked around quite easily I think (e.g. re-commit all images with `manylinux-python ensure-all` called as a fixture).

> Just thinking.... another idea could be a manylinux-slim series of images, which contain cpython only, and a set of manylinux-full images, which use manylinux-slim as a BASE and install all the interpreters on top. No idea how much work it would be to make that happen, but would be quite neat. cibuildwheel could choose the slim variants if it only needs cpython, but use the full ones in the tests.

I don't think it would be too difficult to add full images but I'd rather not multiply images if the "fix" is just to call `manylinux-python ensure-all` or `manylinux-python ensure ...` and the time is similar to pulling the image.

The only issue I can see would be local usage, where, not to re-install everytime, one would need to re-commit all images with `manylinux-python ensure-all` (as proposed for cibuildwheel tests).

docker/build_scripts/python-tag-abi-tag.py
```diff
@@ -2,9 +2,16 @@
...
 
 
 # first tag is always the more specific tag
 tag = next(sys_tags())
+
+# there are some inconsistencies so check against wheel
+wheel_abi_tag = get_abi_tag()
+if tag.abi != wheel_abi_tag:
+    assert tag.interpreter.startswith("graalpy")
+    tag = Tag(tag.interpreter, wheel_abi_tag, tag.platform)
```
mayeut, line 16: @timfel,

It seems that packaging does not agree with wheel w.r.t. to the ABI tag. This is the sign that something bad will happen in the future (I guess when pip will update packaging: https://github.com/pypa/pip/issues/11715).
The difference comes from https://github.com/pypa/packaging/commit/30554f503a5368aabf6ee571bd9cdf29d5f00c12

Unfortunately, the ecosystem did not move forward with a sysconfig like that to express this ABI. xref https://discuss.python.org/t/sysconfig-should-provide-more-information-to-help-packaging/22950 and https://github.com/pypa/packaging/issues/330

tests/run_tests.sh
```diff
@@ -58,12 +94,21 @@ for PYTHON in /opt/python/*/bin/python; do
...
-	${PYTHON} -m pip install ${REPAIRED_WHEEL}
-	if [ "$(${PYTHON} -c 'import forty_two; print(forty_two.answer())')" != "42" ]; then
-		echo "invalid answer, expecting 42"
-		exit 1
+	if [ "${IMPLEMENTATION}" != "graalpy" ]; then
+		${PYTHON} -m pip install ${REPAIRED_WHEEL}
+		if [ "$(${PYTHON} -c 'import forty_two; print(forty_two.answer())')" != "42" ]; then
+			echo "invalid answer, expecting 42"
+			exit 1
+		fi
```
mayeut, line 102: @timfel, the wheel just built can't be installed by pip, there's a mismatch in the platform tag: `linux_aarch64` vs `linux_armv7l`.

mayeut:
```
[root@95647b9a4112 /]# graalpy3.10 -m pip debug
...
Compatible tags: 27
  graalpy310-graalpy230_310_native_aarch64_linux-linux_armv7l
  ...
  [First 10 tags shown. Pass --verbose to show all.]

[root@95647b9a4112 /]# graalpy3.10 -c "from packaging.tags import sys_tags; print(next(sys_tags()))"
graalpy310-graalpy230_310_native-linux_armv7l
```
This time packaging agrees but also gets it wrong. I guess this is related to the `is_32bit` check.

tests/run_tests.sh
```diff
@@ -26,44 +61,55 @@ for PYTHON in /opt/python/*/bin/python; do
...
-		echo "invalid answer, expecting 42"
-		exit 1
+	if [ "${IMPLEMENTATION}" != "graalpy" ]; then
+		SRC_DIR=/tmp/forty-two-${IMPLEMENTATION}${PYVERS}
+		DIST_DIR=/tmp/dist-${IMPLEMENTATION}${PYVERS}
+		cp -rf ${MY_DIR}/forty-two ${SRC_DIR}
+		PY_ABI_TAGS=$(basename $(dirname $(dirname $PYTHON)))
+		EXPECTED_WHEEL_NAME=forty_two-0.1.0-${PY_ABI_TAGS}-linux_${AUDITWHEEL_ARCH}.whl
+		${PYTHON} -m build -w -o ${DIST_DIR} ${SRC_DIR}
+		if [ ! -f ${DIST_DIR}/${EXPECTED_WHEEL_NAME} ]; then
```
mayeut, line 88: On x86_64, the wheel does not have the expected name. The difference in behavior with aarch64 comes from the 32bit check of wheel only implemented for x86_64 but not aarch64

docker/build_scripts/python-tag-abi-tag.py
timfel: We have a PR to wheel to fix the `get_abi_tag` function in `src/wheel/bdist_wheel.py`. When I signed off on the change in packaging, I actually expected pip to upgrade their vendored version quicker. Right now GraalPy is in this in-between state there, with pip having the old vendored packaging, wheel having a copy the logic in [bdist_wheel.py](https://github.com/pypa/wheel/blob/main/src/wheel/bdist_wheel.py#L91), and packages like maturin and meson-python already implementing the new logic. I hope it'll just slowly resolve as the necessary changes that started with that packaging commit slowly perculate through everywhere.

tests/run_tests.sh
timfel: Yes, the `is_32bit` check is biting us. The PR you linked didn't move forward mostly because of concern for 32-bit macOS. We have found making sys.maxsize larger on GraalPy brings other issues though, since some code assumes that (as documented) this gives the maximum array size.

tools/update_python_downloads.py
timfel, line 1: FYI, I added something to get the versions of graalpy to my earlier cibuildwheel PR: https://github.com/timfel/cibuildwheel/commit/ace30fbd13f344f7727dd411ec938d0fb01b2616#diff-2fc4993e0fc3281ad7dfa0e696c6791c9339e16180d62b6fdd05df72a217ced5R112. Not sure if we want to do something like this here to more easily update graalpy, too. I can also do that in a separate PR if/when this one goes in.

docker/build_scripts/python-tag-abi-tag.py
mayeut: xref https://github.com/python/cpython/issues/99560

docker/build_scripts/python_versions.json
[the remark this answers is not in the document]
joerick: cool, even better :) . JSON output would be handy.

docker/build_scripts/manylinux-python.py
```diff
@@ -0,0 +1,168 @@
...
+            install_path = INSTALL_DIR.joinpath(tag)
+            installed = install_path.exists()
+            if installed:
+                info = get_info_from_path(install_path)
+            else:
+                info = get_info_from_tag(tag)
+            print(f"  installed:            {'yes' if installed else 'no'}")
+            print(f"  python version:       {info['pv']}")
+            print(f"  implemention:         {info['i']}")
+            print(f"  implemention version: {info['iv']}")
```
joerick, line 111: Here's an implementation that supports JSON output:

```suggestion
def _list(args):
    tags = get_all_tags()
    if args.installed:
        tags = [tag for tag in tags if INSTALL_DIR.joinpath(tag).exists()]

    tag_infos = []

    for tag in tags:
        install_path = INSTALL_DIR.joinpath(tag)
        installed = install_path.exists()
        if installed:
            info = get_info_from_path(install_path)
        else:
            info = get_info_from_tag(tag)

        tag_info = {
            "identifier": tag,
            "installed": installed,
            "python_version": info["pv"],
            "implementation": info["i"],
            "implementation_version": info["iv"],
            "install_path": install_path,
        }
        tag_infos.append(tag_info)

    if args.format == 'text':
        for tag_info in tag_infos:
            if not args.verbose:
                print(tag_info["identifier"])
            else:
                print(f"{tag_info['identifier']}:")
                print(f"  installed:              {'yes' if tag_info['installed'] else 'no'}")
                print(f"  python version:         {tag_info['python_version']}")
                print(f"  implementation:         {tag_info['implementation']}")
                print(f"  implementation version: {tag_info['implementation_version']}")
    elif args.format == 'json':
        json.dump(tag_infos, sys.stdout, indent=2)
    else:
        raise ValueError(f'unknown format {args.format}')
```

It would also need 

```python
    parser.add_argument("--format", choices=["text", "json"], default="text")
```

in `add_parser_list()`.

docker/build_scripts/python-tag-abi-tag.py
msimacek: Here's the wheel PR: https://github.com/pypa/wheel/pull/557

mayeut:
I'm going to split the PR given there are still some things to workout throughout the ecosystem for GraalPy.

docker/build_scripts/python_versions.json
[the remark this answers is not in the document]
mayeut: done, thanks for your snippet.

docker/build_scripts/python-tag-abi-tag.py
mayeut: marking as resolved in this PR. follow up in #1520

tools/update_python_downloads.py
mayeut: marking as resolved in this PR. follow up in #1520

tests/run_tests.sh
mayeut: marking as resolved in this PR. follow up in https://github.com/pypa/manylinux/pull/1520

mayeut:
There's still some doc missing for manylinux-interpreters, otherwise, I think all comments have been addressed.

@joerick, am I missing something in my answers or do you have more comments ?

joerick:
Sorry for the delay, all good from my perspective. I'm not fully understanding how we make this fast in the cibuildwheel CI context, but that's downstream it sounds like you have an idea :)

PR #1554, opened 2023-12-05
Build with shared libraries

akaszynski:
Build Python 3.10 with shared libraries.


akaszynski:
Sorry! Should have remained on my fork.

PR #1520, opened 2023-08-19
Add GraalPy

mayeut:
In order to install GraalPy in the image, one must call `manylinux-interpreters ensure graalpy310-graalpy230_310_native`.

closes #1509

PR #1564, opened 2023-12-29
Add CPython 3.13 (with & without `--disable-gil`)

### LINKED ISSUE #1548: Support for an experimental version of free-threaded (no-gil) Python 3.13
[PEP 703](https://peps.python.org/pep-0703/) was accepted recently, meaning that free-threaded Python is moving into a new phase, where the goal is to see how it would really work in the Python ecosystem, with real C extensions! A large part of the unknowns of the project lie in how the C extension world will work, when there's no global lock protecting Python objects.

Enabling wheel-building seems like it might help accelerate this kind of testing. I think normally support for interpreters is only added at the beta phase, but perhaps the situation would warrant a special case?

refs: 
- https://github.com/pypa/cibuildwheel/issues/1657 
- https://github.com/pypa/packaging/pull/728 
- https://github.com/python/cpython/issues/111870

mayeut:
fix #1548


mayeut:
@pradyunsg,

follow-up on [your request](https://github.com/pypa/cibuildwheel/issues/1657#issuecomment-2106213958) to add support for free-threaded builds to cibuildwheel:
I was kind of hoping that the release of pip would be any day now.
I  see there are a couple of issues blocking a release for now and, if the release can't be planned in the near future, then the [option proposed](https://github.com/pypa/cibuildwheel/issues/1657#issuecomment-2106268082) by @henryiii to use the pip beta on python 3.13 (or maybe python 3.13 free-threaded only) could be implemented here.

henryiii:
I think it would be pretty reasonable to use the beta on 3.13 only. I don't think there's an environment selector for free threaded? I'd prefer not to differentiate on free threaded if there's no environment selector for it.

henryiii:
Why is setuptools/wheel on 3.12, actually?

mayeut:
As mentioned in the commit message, it shouldn't be there & perhaps especially for python 3.12 where it's not part of `ensurepip` / `venv`. Keeping it only as a precaution not to introduce regressions downstream.

henryiii:
I didn't see the description, good. I expect after 3.13 is out for a while, it probably can be removed safely from 3.12.

pradyunsg:
> I was kind of hoping that the release of pip would be any day now.

I'd have liked that too but we also have the other part of the packaging upgrade affecting us in this release: https://pradyunsg.me/blog/2024/05/13/pip-24-1-betas/. Due to some non-public health issues on my part, the actual process and comms around this haven't gone through as smoothly/effectively as I'd like. 😅

mayeut:
Thanks for the feedback @pradyunsg !

I'll merge without waiting for the final release then.

FYI, pip 24.1b1 can install properly free-threaded wheels now (there's a minimal build / install test here). It also fixes issues with graalpy (c.f. #1520) but this one will have to wait for the final release.

henryiii:
In cause people are trying this out, `python3.13t -m venv ...` still installs pip 24.0, something to watch for when testing this! Remember to upgrade pip in venvs.

henryiii:
Ouch, this includes isolated builds.

PR #1570, opened 2024-02-02
Build shared libpython.so

lpsinger:
Our project (http://git.ligo.org/lscsoft/lalsuite, https://pypi.org/project/lalsuite/) distributes several C command line programs that embed Python interpreters by linking against libpython, plus some Python C extensions modules (which of course do not link against libpython).

For those rare cases like ours where Python embedding is required, using a shared library libpython permits auditwheel to deduplicate libpython and store only one copy in the wheel.


mayeut:
Thanks for your contribution.

I understand your needs which are different from the past ones to get a shared build in this repo however, this adds too many risks which I'm not prepared to deal with.

As I've said before, I'm not against getting support to build a shared version as long as it's an optional non-default feature. The images would need to be generated / published elsewhere though.

lpsinger:
That makes sense. Given that we have a valid application, would you view #1185 favorably?

Issue #1574, opened 2024-02-13
Error Repairing Wheel to manylinux2014_x86_64 ABI Due to Too-Recent Versioned Symbols


cyprienc:
When attempting to repair the wheel for `triton_nightly` to the `manylinux2014_x86_64` ABI, the auditwheel tool fails with the following error:

```
auditwheel: error: cannot repair "/tmp/cibuildwheel/built_wheel/triton_nightly-3.0.0.post20240212030344-cp38-cp38-linux_x86_64.whl" to "manylinux2014_x86_64" ABI because of the presence of too-recent versioned symbols. You'll need to compile the wheel on an older toolchain.
```
This issue occurs during the GitHub Actions workflow defined [here](https://github.com/openai/triton/blob/54220c010fb2b0a3e6a9ce4b3ddaa2fe5704a035/.github/workflows/wheels.yml) (error started appearing around Jan 11, 24).

An example of a failed run can be found [here](https://github.com/openai/triton/actions/runs/7722889468/job/21051879669#logs).

# Environment
CI/CD Platform: GitHub Actions
Container Image Used: `quay.io/pypa/manylinux2014_x86_64:latest`
Local Testing Command (comment out the Azure steps): `act -P CPU=ghcr.io/catthehacker/ubuntu:act-22.04 -W .github/workflows/wheels.yml workflow_dispatch`

# Questions
Why are the symbols not compatible with `manylinux2014` despite using the specified base image for the build environment?
Is there a recommended approach to ensure compatibility with `manylinux2014` ABI standards when building wheels in this environment?

(This is related to https://github.com/openai/triton/issues/2616)

cyprienc:
Here is the (large) dump of the output of `auditwheel` during repair:
```
[Wheels/Build-Wheels]   ❓  ::group::Repairing wheel...
|
|     + sh -c 'auditwheel --verbose repair -w /tmp/cibuildwheel/repaired_wheel /tmp/cibuildwheel/built_wheel/triton_nightly-3.0.0.post20240212030344-cp38-cp38-linux_x86_64.whl'
| DEBUG:auditwheel.musllinux:musl libc not detected
| DEBUG:auditwheel.libc:Falling back to GNU libc
| INFO:auditwheel.main_repair:Repairing triton_nightly-3.0.0.post20240212030344-cp38-cp38-linux_x86_64.whl
| DEBUG:auditwheel.wheel_abi:processing: triton/_C/libtriton.so
[...]

PR #1575, opened 2024-02-14
feat: manylinux_2_34 based on AlmaLinux 9

cyprienc:
Hi, 

I imagine this will require more discussions but I thought I'd at least open an MR proposing a `manylinux_2_34` compliant image based on AlmaLinux 9. 

I need it to build the nightly wheels of [openai/triton](https://github.com/openai/triton) which are currently failing because the nvidia and amd backends are now built using more recent versions of stdlibc++ and co. 

This relates to https://github.com/openai/triton/issues/2616 and https://github.com/pypa/manylinux/issues/1574. 

Thanks !


docker/Dockerfile
```diff
@@ -1,6 +1,6 @@
 # default to latest supported policy, x86_64
-ARG BASEIMAGE=amd64/almalinux:8
-ARG POLICY=manylinux_2_28
+ARG BASEIMAGE=amd64/almalinux:9
```
dralley, line 2: I was under the impression from [discussion](https://github.com/pypa/manylinux/issues/1282#issuecomment-1548916560) on previous issues that manylinux_2_34 would (might?) use the UBI or UBI-minimal image, though I don't think there was ever an actual conclusion reached on the subject.

https://catalog.redhat.com/software/containers/ubi9/ubi/615bcf606feffc5384e8452e?architecture=amd64&image=65a82982a10f3e68777870b5

https://catalog.redhat.com/software/containers/ubi9/ubi-minimal/615bd9b4075b022acc111bf5?architecture=amd64&image=65a828e3cda4984705d45d26

@carlwgeorge @mattip @tiran @h-vetinari @kpfleming You were all active on the previous discussion, any input?

docker/build_scripts/install-runtime-packages.sh
```diff
@@ -32,7 +32,7 @@ source $MY_DIR/build_utils.sh
 
 
 # MANYLINUX_DEPS: Install development packages (except for libgcc which is provided by gcc install)
-if [ "${AUDITWHEEL_POLICY}" == "manylinux2014" ] || [ "${AUDITWHEEL_POLICY}" == "manylinux_2_28" ]; then
+if [ "${AUDITWHEEL_POLICY}" == "manylinux2014" ] || [ "${AUDITWHEEL_POLICY}" == "manylinux_2_28" ] || [ "${AUDITWHEEL_POLICY}" == "manylinux_2_34" ]; then
 	MANYLINUX_DEPS="glibc-devel libstdc++-devel glib2-devel libX11-devel libXext-devel libXrender-devel mesa-libGL-devel libICE-devel libSM-devel zlib-devel expat-devel"
```
dralley, line 36: `libX11-devel`, `libXrender-devel`, `mesa-libGL-devel` - do we know for certain that these still need to be carried forwards?

docker/build_scripts/install-runtime-packages.sh
```diff
@@ -97,7 +97,11 @@ elif [ "${AUDITWHEEL_POLICY}" == "manylinux_2_28" ]; then
 	sed -i '/^override_install_langs=/d' /etc/yum.conf
 	dnf -y upgrade
 	dnf -y install dnf-plugins-core
-	dnf config-manager --set-enabled powertools # for yasm
+	if [ "${AUDITWHEEL_POLICY}" == "manylinux_2_34" ]; then
+		dnf config-manager --set-enabled crb
+	else
+		dnf config-manager --set-enabled powertools # for yasm
+	fi
```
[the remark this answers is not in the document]
dralley: Which probably explains the CI failure

> Error: Unable to find a match: yasm libidn

docker/build_scripts/update-system-packages.sh
```diff
@@ -20,7 +20,7 @@ if [ "${AUDITWHEEL_POLICY}" == "manylinux2014" ]; then
 	fi
 	yum clean all
 	rm -rf /var/cache/yum
-elif [ "${AUDITWHEEL_POLICY}" == "manylinux_2_28" ]; then
+elif [ "${AUDITWHEEL_POLICY}" == "manylinux_2_28" ] || [ "${AUDITWHEEL_POLICY}" == "manylinux_2_34" ]; then
 	dnf -y upgrade
 	dnf clean all
 	rm -rf /var/cache/yum
```
dralley, line 26: This instruction is incorrect even for `manylinux_2_28`, `/var/cache/yum/` is now `/var/cache/dnf/` on systems that use dnf as the package manager.

docker/build_scripts/install-runtime-packages.sh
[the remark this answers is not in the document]
dralley: And if we enable the repo for `manylinux_2_34`, we potentially ought to enable it for `manylinux_2_28`

https://github.com/pypa/manylinux/issues/1338

docker/Dockerfile
carlwgeorge: I would love to see manylinux_2_34 use UBI 9 as a base.  UBI is a subset of RHEL packages that are freely redistributable, and RHEL is what Alma aspires to be compatible with.  It's better to just go straight to the primary thing in my opinion.  If there are packages missing from the UBI package set that would block this, we can request those be added (see question 34 [here](https://developers.redhat.com/articles/ubi-faq#support__lifecycle__and_updating-h2)).

docker/build_scripts/install-runtime-packages.sh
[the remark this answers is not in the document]
carlwgeorge: I'll also note that UBI9 has CRB enabled by default, so if a packages are needed from there it doesn't need an explicit enablement step.

docker/Dockerfile
h-vetinari: There really should be a separate discussion issue on `manylinux_2_34`, separate from #1574, to discuss choices/timelines/details around the implementation, with the necessary context, plus referencing the `_2_28` meta-issue and previous discussions ([example](https://discuss.python.org/t/the-next-manylinux-specification/1043/239)) on `_2_34`.

h-vetinari: > There really needs to be a discussion issue on `manylinux_2_34`

https://github.com/pypa/manylinux/issues/1585

docker/build_scripts/install-runtime-packages.sh
mayeut: It's better to address this comment in a separate issue / follow-up PR.
Indeed, some might not be relevant anymore since PEP600 but kept for an easy transition.

docker/Dockerfile
mayeut: tracking issue opened, resolving conversation here.

docker/build_scripts/install-runtime-packages.sh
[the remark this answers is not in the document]
mayeut: épelasses now enabled.

EwoutH:
Nice milestone: glibc 2.34 is now [supported](https://mayeut.github.io/manylinux-timeline/) over 50% of consumer systems using Python.

image

glibc 2.35 is also supported on more than half of the systems, probably since Ubuntu 22.04 has glibc 2.35 bundled by default (and has a huge marketshare).

README.rst
```diff
@@ -95,6 +97,24 @@ done the work for you. The images are uploaded to `quay.io`_ and are tagged
 for repeatable builds.
 
 
+manylinux_2_34 (AlmaLinux 9 based)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Toolchain: GCC 13
```
EwoutH, line 103: If I understood https://github.com/pypa/manylinux/issues/1585#issuecomment-2198069688 correctly, the request from @mayeut is to use the new GCC 14 toolset for `manylinux_2_34`.

mayeut:
Thanks for your initial work on this @cyprienc 
I'll close this PR and follow-up in #1698

PR #1590, opened 2024-03-22
chore: use uv instead of pip-tools

henryiii:
This moves the lockfile generation from pip-tools to uv. The update_python_dependencies step now runs in under 2 seconds vs. ~40 seconds, and doesn't require all supported versions of Python to be installed (not sure about this, investigating). The update_python_tools job now takes under 2 seconds vs. 17 seconds, and can be run from any Python version. (I'm also using NOX_DEFAULT_VENV_BACKEND=uv locally for the uv job timings)

This stops updating the lock file for 3.6, as nothing should be updating for it anymore.


.github/workflows/update-dependencies.yml
```diff
@@ -11,14 +11,10 @@ on:
 jobs:
   update-dependencies:
     name: Update dependencies
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
```
mayeut, line 14:
```suggestion
    runs-on: ubuntu-latest
```

it was pinned to 20.04 for python 3.6 and can be moved to latest now.

docker/build_scripts/requirements3.10.txt
```diff
@@ -1,23 +1,26 @@
...
-#
-#    nox -s update_python_dependencies-3.10
-#
+# This file was autogenerated by uv via the following command:
+#    nox -s update_python_dependencies
 build==1.1.1 \
     --hash=sha256:8ed0851ee76e6e38adce47e4bee3b51c771d86c64cf578d0c2245567ee200e73 \
     --hash=sha256:8eea65bb45b1aac2e734ba2cc8dad3a6d97d97901a395bd0ed3e7b46953d2a31
-    # via -r requirements.in
+importlib-metadata==7.1.0 \
```
mayeut, line 6: This seems strange. Why does pip-compile and uv compile behavior differ ?
importlib-metadata was not required before and as far as I can see, build version did not change.

henryiii:
I was a little worried about the warning that 3.11 was being used instead of 3.7 since it wasn't installed, but it's resolving for 3.7, so I guess it's harmless.

docker/build_scripts/requirements3.10.txt
henryiii: https://github.com/pypa/build/blob/b36126a2b7a387757519029d121da17a38135d67/pyproject.toml#L41


Looks like uv resolves for 3.10.0, while pip-compile uses whatever it was running on.

mayeut:
took me quite a while to find time to get to this.
thanks Henry

PR #1614, opened 2024-05-14
feat: add uv (except musl s390x)

### LINKED ISSUE #1607: Include uv in tools?
I'd like to try uv for cibuildwheel, and it's a single binary that works across Python versions, so it seems like adding it here would be a good first step. Thoughts?

Also, uv already supports free-threaded builds since https://github.com/astral-sh/uv/pull/2805. :)

henryiii:
This adds uv to all the images currently supported. ~~I think musl powerpc64le might be possible in the near future (edit: https://github.com/astral-sh/uv/pull/3537)~~ (Edit: now available). This also updates the README with details, and speeds up the noxfile and uses PEP 723 at the expense of requiring a recent nox.

Only musl s390x missing now.

Close #1607.


henryiii:
Looks like we are stuck in a libtool release: https://lists.gnu.org/archive/html/autotools-announce/2024-05/msg00000.html

mayeut:
Thanks for the doc & nox updates also.

PR #1689, opened 2024-10-09
Update downloaded interpreters to add GraalPy 24.1

timfel:
- updating graalpy311-graalpy241_311_native x86_64 to 24.1.0
- updating graalpy311-graalpy241_311_native aarch64 to 24.1.0

### DIFF
--- a/docker/build_scripts/python_versions.json
+++ b/docker/build_scripts/python_versions.json
@@ -1,14 +1,14 @@
 {
-  "graalpy310-graalpy240_310_native": {
+  "graalpy311-graalpy241_311_native": {
     "x86_64": {
-      "version": "24.0.2",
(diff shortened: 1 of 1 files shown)


timfel:
@mayeut How would like me to handle the expected interpreter count? Should I increase it by 1, or should I remove the 24.0 graalpy release and only keep the latest? Either way is fine from my perspective, although maybe keeping the current and the last graalpy release might be more convenient for projects

timfel:
I decided to drop 24.0 right away, cibuildwheel needs 24.1 to work and the error message for users is pretty clear (says it cannot find graalpy310 and will list what is available)

auvipy (commented):
started the CI

Issue #1585, opened 2024-03-14
Tracking issue for manylinux_2_34 image


h-vetinari:
As time goes by, projects start using new glibc features not present in old versions, hence requiring a newer manylinux baseline version to be defined.

For example:
* https://github.com/openai/triton/issues/2616 
* https://github.com/pypa/manylinux/issues/1574

Also, according to the [manylinux-timeline](https://mayeut.github.io/manylinux-timeline/), around ~40% of users on non-EOL Python versions have better than glibc 2.34 already.

IMO the discussion in #1282 (and the failure of Debian-based `2_24`, c.f. #1012) showed comprehensively that we should stick with a RHEL-derivative also for the next manylinux standard - the devtoolset backports make all the difference w.r.t. to longevity, and all those flavours are ABI compatible. 

This would also continue the pattern for all major manylinux standards so far (again excepting `2_24`):

| manylinux | glibc | RHEL |
| -- | -- | -- |
manylinux1 | 2.5 | 5
manylinux2010 | 2.12 | 6
manylinux2014 | 2.17 | 7
manylinux_2_28 | 2.28 | 8
manylinux_2_34 | 2.34 | 9

Updating the flavour table from #1282:

> The big question is what base image to use. We can choose between C9S (`centos:stream9`), AlmaLinux, RockyLinux, or UBI. Personally I would prefer an image based entirely on open source software. [...]

| base image | EOL¹ | gcc-toolset-13² | aarch64 | ppc64le | s390x | bug fix availability |
|-|-|-|-|-|-|-|
| CentOS Stream 9 | 2027-05-31³ | yes | yes³ | yes³ | yes³ | 1st |
| UBI 9 | 2032-05-31 | yes⁴ | yes⁵ | yes⁵ | yes⁵ | 2nd |
| AlmaLinux 9 | 2032-05-31 | yes | yes⁶ | yes⁶ | yes⁶ | 3rd |
| RockyLinux 9 | 2032-05-31 | yes | yes⁷ | yes⁷ | yes⁷ | 3rd | 

<sup>1: including Maintenance Support, https://access.redhat.com/support/policy/updates/errata/#Life_Cycle_Dates</sup>
<sup>2: https://pkgs.org/download/gcc-toolset-13</sup>
<sup>3: https://www.centos.org/centos-stream/</sup>
<sup>4: [at least](https://github.com/pypa/manylinux/issues/1282#issuecomment-1548509593) gcc 12, probably on gcc 13 already</sup>
[...]

PR #1698, opened 2024-10-27
feat: Add manylinux_2_34 image

### LINKED ISSUE #1338: Enable EPEL repo for `manylinux_2_28`
`manylinux_2_28` images are not configured to install dependencies such as `msgpack` that are only available in EPEL repositories. Enabling EPEL will reduce the amount of Linux-fu that users need to create binary wheels.

mayeut:
towards #1585
close #1575
fix #1338

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -37,6 +37,8 @@ The manylinux project supports:
 
 - ``manylinux_2_28`` images for ``x86_64``, ``aarch64``, ``ppc64le`` and ``s390x``.
 
+- ``manylinux_2_34`` images for ``x86_64``, ``aarch64``, ``ppc64le`` and ``s390x``.
+
 - ``musllinux_1_2`` images for ``x86_64``, ``i686``, ``aarch64``, ``ppc64le``, ``s390x`` and ``armv7l``.
 
 
@@ -91,6 +93,24 @@ Rather than forcing you to install an old distro yourself, install Python,
 etc., we provide `Docker <https://docker.com/>`_ images where we've
 done the work for you. The images are uploaded to `quay.io`_ and are tagged
 for repeatable builds.
+
+manylinux_2_34 (AlmaLinux 9 based)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Toolchain: GCC 14
+
+- x86_64 image: ``quay.io/pypa/manylinux_2_34_x86_64``
+- aarch64 image: ``quay.io/pypa/manylinux_2_34_aarch64``
+- ppc64le image: ``quay.io/pypa/manylinux_2_34_ppc64le``
+- s390x image: ``quay.io/pypa/manylinux_2_34_s390x``
+
+Built wheels are also expected to be compatible with other
+distros using glibc 2.34 or later, including:
+
+- Debian 12+
+- Ubuntu 21.10+
+- Fedora 35+
+- CentOS/RHEL 9+
 
 
 manylinux_2_28 (AlmaLinux 8 based)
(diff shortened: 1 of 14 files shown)


README.rst
```diff
@@ -92,6 +94,24 @@ etc., we provide `Docker <https://docker.com/>`_ images where we've
 done the work for you. The images are uploaded to `quay.io`_ and are tagged
 for repeatable builds.
 
+manylinux_2_34 (AlmaLinux 9 based)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Toolchain: GCC 13
```
h-vetinari, line 100: Didn't you say in #1585 that you'd want to start off with GCC 14?

build.sh
```diff
@@ -50,12 +44,22 @@ if [ "${POLICY}" == "manylinux2014" ]; then
...
-	BASEIMAGE="${MULTIARCH_PREFIX}almalinux:8"
+	BASEIMAGE="almalinux:8"
 	DEVTOOLSET_ROOTPATH="/opt/rh/gcc-toolset-12/root"
 	PREPEND_PATH="${DEVTOOLSET_ROOTPATH}/usr/bin:"
 	LD_LIBRARY_PATH_ARG="${DEVTOOLSET_ROOTPATH}/usr/lib64:${DEVTOOLSET_ROOTPATH}/usr/lib:${DEVTOOLSET_ROOTPATH}/usr/lib64/dyninst:${DEVTOOLSET_ROOTPATH}/usr/lib/dyninst"
+elif [ "${POLICY}" == "manylinux_2_31" ]; then
+	BASEIMAGE="ubuntu:20.04"
+	DEVTOOLSET_ROOTPATH=
+	PREPEND_PATH=
+	LD_LIBRARY_PATH_ARG=
```
h-vetinari: This commit seems to mix in some changes for 2_31 as well? Should probably be a separate change/PR?

mayeut: The plan is already to split this before setting out of draft.

README.rst
mayeut: I said that "I'd like to wait" but if I don't get some sort of planning/roadmap for availability (it's been in stream 9 for a while now but the fact that it's nowhere else requires more investigation on why), we might want to start with 13 & update later on.

mayeut: Digging a bit more into this I only could find that there's some on-going work in RHEL bug tracker. It seems they're already using it in a staging area but still finding bugs and reworking if I understand correctly. I have not found any kind of planning around this.

h-vetinari: Makes sense. 👌

We're also using GCC 13 in conda-forge until GCC 15 is released, for pretty much those reasons (picking up all the bugfixes over the course of a year leading up to the next major version).

carlwgeorge: I can't speak to the Alma roadmap, but in general they try to match RHEL.  RHEL doesn't make exact dates public, but there is a public [planning guide](https://access.redhat.com/support/policy/updates/errata#RHEL9_Planning_Guide) that should give you a rough idea.  New packages such as GCC 14 are typically delivered in CentOS Stream once they pass QA, and then show up in the next RHEL minor version.  Those minor versions happen about every six months (since RHEL 8, before that it was not consistent).  So the next time you see a feature in CentOS Stream and want to know when it will be in RHEL, look back at the [release date](https://access.redhat.com/articles/3078) for the last minor version, and estimate six months after that.

mayeut: 9.5 has been released and includes gcc-toolset-14

h-vetinari:
Is there anything concrete that this PR is still waiting for? :)

mayeut:
> Is there anything concrete that this PR is still waiting for? :)

Per https://github.com/pypa/manylinux/issues/1585#issuecomment-2481320583, It was planned for the end of the week if there were no more reviews delaying this. It's the end of the week and there are no more reviews delaying this => merging now.

PR #1747, opened 2025-02-11
chore: use sigstore to verify CPython archives

mayeut:
CPython 3.14 won't include an OpenPGP signature.
Let's move to sigstore for CPython 3.7+


auvipy (approved):
yeah, the change seems to be inevitable

PR #1810, opened 2025-07-10
fix: enable missing extensions on sqlite3

### LINKED ISSUE #1809: sqlite3 missing FTS5 extension
The issue is described here: https://github.com/marcelotduarte/cx_Freeze/issues/3038

I immediately realized it was an issue with manylinux, so I did some testing and discovered the following:
- cx_Freeze 7.2.10 has sqlite3 3.47.0
- cx_Freeze 8.0 has version 3.49.1

I found that what may have changed between these versions is that sqlite3 used autoconf and now uses [autosetup](https://msteveb.github.io/autosetup/), which may have disabled the inclusion of extensions. autosetup has a wrapper for autoconf.

Here's a little test to see which extensions are enabled:
```python
from pprint import pprint
import sqlite3

con = sqlite3.connect(":memory:")
cur = con.cursor()
cur.execute("pragma compile_options;")
available_pragmas = cur.fetchall()
con.close()

print(sqlite3.__name__, sqlite3.sqlite_version)
pprint(available_pragmas)
```

marcelotduarte:
Closes #1809 
I believe this is the solution, from what I read in https://sqlite.org/src/doc/trunk/README.md

### DIFF
--- a/docker/tests/modules-check.py
+++ b/docker/tests/modules-check.py
@@ -10,6 +10,11 @@ class TestModules(unittest.TestCase):
         print(f"{sqlite3.sqlite_version=}", end=" ", flush=True)
         assert sqlite3.sqlite_version_info[0:2] >= (3, 50)
 
+        # When the extension is not installed, it raises:
+        # sqlite3.OperationalError: no such module: fts5
+        conn = sqlite3.connect(":memory:")
+        conn.execute("create virtual table fts5test using fts5 (data);")
+
     def test_tkinter(self):
         # Make sure tkinter module can be loaded properly
         import tkinter as tk


marcelotduarte:
The patch added:
```
('ENABLE_FTS3',),
 ('ENABLE_FTS4',),
 ('ENABLE_FTS5',),
 ('ENABLE_GEOPOLY',),
 ('ENABLE_PREUPDATE_HOOK',),
 ('ENABLE_RTREE',),
 ('ENABLE_SESSION',),
```

mayeut (changes requested):
Thanks for the patch @marcelotduarte.
Can a check be added in `docker/tests/modules-check.py` to make sure no regressions are introduced in the future ?

PR #1944, opened 2026-05-18
Support newer GraalPy release assets that include the Python version.

timfel:
With https://github.com/oracle/graalpython/pull/810 we will slightly alter upcoming GraalPy release artifact naming. The name will now include the compatible CPython version name in the filename. This change adapts the interpreter downloads script to support both the older and newer format. We can drop the older one once support for GraalPy versions older than 25.1 is removed.

/cc @msimacek
Related:  https://github.com/oracle/graalpython/issues/623

### DIFF
--- a/tools/update_interpreters_download.py
+++ b/tools/update_interpreters_download.py
@@ -89,7 +89,10 @@ def update_pypy_versions(versions, updates):
 
 
 def update_graalpy_version(releases, graalpy_spec, tag, arch, version_dict, updates):
-    graalpy_arch = {"x86_64": "amd64"}.get(arch, arch)
+    graalpy_arch = re.escape({"x86_64": "amd64"}.get(arch, arch))
+    asset_re = re.compile(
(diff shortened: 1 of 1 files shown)


mayeut (changes requested):
@timfel,

Thanks for the PR.

The architecture should be part of the pattern (re escaped to be on the safe side) since we know what to look for, no need to capture.
The other captured group should be named to ease readability.

Given the PR I assume there will still only be a single minor version of python for any given minor version of graalpy. If that assumption is not correct anymore, we're missing a check on the python version.

timfel:
> The architecture should be part of the pattern (re escaped to be on the safe side) since we know what to look for, no need to capture.
> The other captured group should be named to ease readability.

Done.

> Given the PR I assume there will still only be a single minor version of python for any given minor version of graalpy.

Correct.

mayeut (approved):

Issue #1878, opened 2025-11-28
Update images to gcc 15 toolchain [postponed]


mayeut:
The following images will be updated to use gcc 15 on 2026-01-03

- manylinux_2_28 (gcc-toolset-14 -> gcc-toolset-15)
- manylinux_2_34 (gcc-toolset-14 -> gcc-toolset-15)
- manylinux_2_29 (gcc-toolset-14 -> gcc-toolset-15)
- musllinux_1_2 (Alpine Linux 3.22 -> Alpine Linux 3.23)

mayeut:
Issues were found after #1873 was merged. This has been reverted in #1900.
The update is to be planned at a later date once issues discovered by auditwheel tests are fixed.

PR #1956, opened 2026-06-05
feat: use gcc-16 when possible

lpsinger:
Just in case 🙏  the issues with patchelf and GCC 16 were fixed in GCC 16. See #1878.


mayeut:
the issues with patchelf likely won't be fixed by upgrading to gcc-16 given it's feature - i.e. not a bug - from recent binutils that revealed a bug in patchelf.
If my information is not up-to-date or my interpretation invalid, please provide links to the fixes.
Diving through the links from here to auditwheel to patchelf, the relevant ones are I think:
- https://github.com/NixOS/patchelf/issues/568
- https://github.com/NixOS/patchelf/issues/639

lpsinger:
Not to mention that gcc-toolset-16 is not yet available.

PR #1970, opened 2026-07-13
Update hardening compiler flags to prevent breakage on systems where Shadow Stack is enabled

sicherha:
Context: Fedora Linux is considering enabling the Shadow Stack feature
of modern x86_64 CPUs for user-space processes
(https://fedoraproject.org/wiki/Changes/ShadowStack).
Shadow-Stack-enabled processes will refuse to `dlopen()` a shared
library that was built without the `-fcf-protection` compiler option.
This will affect binary Python extensions.

Seize the opportunity to refresh the hardening compiler flags, which
were previously based on Ubuntu 20.04's defaults.

Use the output of `dpkg-buildflags --export` on Ubuntu 24.04 with the
following feature selection:
```
hardening=+all optimize=-lto qa=-framepointer
reproducible=-fixdebugpath,-fixfilepath
```

List of changes:
* Increase `_FORTIFY_SOURCE` level to 3
* Remove `-Wall` (was this previously set manually?)
* Remove mapping of debug-info prefixes (didn't seem to make much sense)
* Add stack-clash protection
* Add branch protection dependent on the CPU architecture

### DIFF
--- a/docker/build_scripts/build-cpython.sh
+++ b/docker/build_scripts/build-cpython.sh
@@ -36,7 +36,7 @@ pushd "Python-${CPYTHON_VERSION}"
 PREFIX="/opt/_internal/cpython-${CPYTHON_VERSION}"
 mkdir -p "${PREFIX}/lib"
 
-CFLAGS_NODIST="${MANYLINUX_CFLAGS} ${MANYLINUX_CPPFLAGS}"
+CFLAGS_NODIST=""
 LDFLAGS_EXTRA=""
 CONFIGURE_ARGS=(--disable-shared --with-ensurepip=no)
 
@@ -115,13 +115,15 @@ fi
 
 unset _PYTHON_HOST_PLATFORM
 
-# configure with hardening options only for the interpreter & stdlib C extensions
-# do not change the default for user built extension (yet?)
+# configure with hardening options for the interpreter & stdlib C extensions and
+# for user built extension
 ./configure \
 	CC=gcc \
 	CXX=g++ \
+	CFLAGS="${MANYLINUX_CFLAGS} ${MANYLINUX_CPPFLAGS}" \
 	CFLAGS_NODIST="${CFLAGS_NODIST}" \
-	LDFLAGS_NODIST="${MANYLINUX_LDFLAGS} ${LDFLAGS_EXTRA}" \
+	LDFLAGS="${MANYLINUX_LDFLAGS}" \
+	LDFLAGS_NODIST="${LDFLAGS_EXTRA}" \
 	"--prefix=${PREFIX}" "${CONFIGURE_ARGS[@]}" > /dev/null
 make > /dev/null
 make install > /dev/null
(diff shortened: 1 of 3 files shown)


mayeut:
approving workflow to seer failures which might be of interest in any case.
please note that those flags do not apply to built wheels so will be useless for the intended purpose, xref #1009 / #59

sicherha:
Looking at the build failures, `-mno-omit-leaf-frame-pointer` is not recognised by the compiler on armv7l; only on aarch and x86_64. In addition, s390x does not seem to treat `-fno-omit-frame-pointer` the same as the other architectures do; Ubuntu sets `-mbackchain` there instead.
Given that frame pointers aren't *that* essential for the use case at hand, I tend towards just leaving the setting off.

Is there a process for rebuilding existing wheels, or do changes to the compiler flags affect only newly submitted package versions?

Either way, even though this pull request will not magically make existing Python extensions compatible with a Shadow-Stack-enabled system, it still makes sense to get on the right track sooner rather than later.

sicherha:
I just saw the relevant part of #1009:
> This does not affect the wheels that are produced by end users as proposed in #59 but mitigates potential security issues in the tools used by manylinux images as mentioned in #1005

Does this mean we cannot influence the flags with which wheels get built? I somehow believed they were built with the same flags as the CPython installation (which are recorded in `sysconfig` and can be obtained with e.g. `sysconfig.get_config_var('CFLAGS')`).

mayeut:
> Does this mean we cannot influence the flags with which wheels get built? I somehow believed they were built with the same flags as the CPython installation (which are recorded in sysconfig and can be obtained with e.g. sysconfig.get_config_var('CFLAGS')).

We can but that must be handled in `docker/build_scripts/build-cpython.sh`.

Can you please post in https://discuss.python.org/c/packaging/14 to get a broader audience for this, the fact is it'll probably impact performances in some way and that package maintainers shall be aware of this change before it goes in (but strictly per PEP600, this change shall go in if Fedora applies this change, which seems uncertain at the moment).

sicherha:
Discussion: https://discuss.python.org/t/building-extensions-with-hardening-flags-via-manylinux/108354

rgommers:
> Does this mean we cannot influence the flags with which wheels get built? 

If the goal is to get 90% of them to add a flag, then yes probably you can throw in some `export CFLAGS`'s in `manylinux`. But not all wheels get built in manylinux containers, and not all `.so`'s in every wheel can be influenced by a simple `export CFLAGS`. So if you need 100%, I think that's probably a lost cause.

> I somehow believed they were built with the same flags as the CPython installation

This is not the case. 

Easy to verify yourself by running for example `python3-config --cflags` and checking if all those flags appear in build logs of Python packages. I just did that with the first build env I had at hand, for `scipy`. `python3-config` output contains `-fno-strict-overflow -Wsign-compare` (matching `sysconfig.get_config_var('CFLAGS')` contents), and the `scipy` build log does not contain those flags. I wouldn't expect any package using `meson-python`, `scikit-build-core` or `maturin` to have matching flags.

